PR #9, opened 2025-06-20
chore: add build task for idls

guibescos:


packages/sessions-idls/.gitignore
```diff
@@ -0,0 +1,2 @@
+src/idl
+src/types
```
guibescos, line 2: i am not very satisfied with the boundary between generated artifacts and not generated here

packages/sessions-idls/turbo.json
```diff
@@ -0,0 +1,16 @@
+{
```
guibescos, line 1: only sessions-idls depends on this

PR #13, opened 2025-06-25
feat: replace nonce by human-readable expiration 

guibescos:


apps/sessions-demo/src/components/Home/enable-trading-button.tsx
```diff
@@ -45,13 +45,15 @@ const handleEnableTrading = async (
   const provider = sessionManagerProgram.provider;
   const sessionKey = Keypair.generate();
 
+  const expires = new Date().toISOString();
```
guibescos, line 48: @cprussin is there a good way to get the ISO string with timezone information?

jayantk (approved):

cprussin (approved):

PR #17, opened 2025-06-26
fix: turbo race condition

guibescos:
There's a race condition between `@fogo/sessions-demo#build` and `@fogo/sessions-demo#test:types` because test:types checks some types inside `.next`

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -160,7 +160,7 @@
       "dependsOn": ["//#install:modules", "^build"]
     },
     "test:types": {
-      "dependsOn": ["//#install:modules", "^build", "build:idl", "build"]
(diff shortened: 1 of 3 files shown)


turbo.json
```diff
@@ -160,7 +160,7 @@
       "dependsOn": ["//#install:modules", "^build"]
     },
     "test:types": {
-      "dependsOn": ["//#install:modules", "^build", "build:idl"]
+      "dependsOn": ["//#install:modules", "^build", "build:idl", "build"]
```
cprussin, line 163: You might want to scope this only to the sessions-demo package, since most packages don't need to be built to type check and enforcing this will slow things down

PR #18, opened 2025-06-26
feat: load alt only once

guibescos:


apps/sessions-demo/src/hooks/use-address-lookup-table.ts
```diff
@@ -0,0 +1,51 @@
...
+  addressLookupTableAddress: string | undefined,
+): State => {
+  const { connection } = useConnection();
+
+  const [state, setState] = useState<State>(State.NotStarted());
+
+  useEffect(() => {
+    if (!addressLookupTableAddress) {
+      setState(State.Complete(undefined));
+    } else if (state.type != StateType.Running) {
```
cprussin, line 15: Do you need to check that the connection is established here?

apps/sessions-demo/src/components/Home/enable-trading-button.tsx
```diff
@@ -136,15 +143,19 @@ export const EnableTradingButton = ({
   const { publicKey, signMessage } = useWallet();
 
   const onEnableTrading = useCallback(() => {
-    if (signMessage && publicKey) {
+    if (
+      signMessage &&
+      publicKey &&
+      addressLookupTableState.type === AccountLookupTableStateType.Complete
```
cprussin, line 149: Generally I prefer to avoid checks like this by doing something like this:

```tsx
const MyComponent = (
  addressLookupTableState: AddressLookupTableState
) => {
  const { publicKey, signMessage } = useWallet();

  return publicKey !== undefined && signMessage !== undefined && addressLookupTableState.type === AccountLookupTableStateType.Complete
    ? <MyComponentImpl publicKey={publicKey} signMessage={signMessage} addressLookupTable={addressLookupTableState.addressLookupTable} />
    : <div>loading...</div>;
}

const MyComponentImpl = (
  publicKey: PublicKey,
  signMessage: (message: Uint8Array) => Promise<Uint8Array>,
  addressLookupTable: AddressLookupTableAccount
) => {
  const onEnableTrading = useCallback(() => {
    // ... here you are guaranteed that `publicKey`, `signMessage`, and `addressLookupTable` are not undefined
  }, [...])
}
```

cprussin (approved):
Nice, LGTM, minor suggestions

PR #24, opened 2025-06-27
feat: directly export typed idls & programs from sessions-idls

cprussin:
This slight improvement means that developers importing the package don't have to do any unsafe typecasting

### DIFF
--- a/apps/sessions-demo/src/components/Home/enable-trading-button.tsx
+++ b/apps/sessions-demo/src/components/Home/enable-trading-button.tsx
@@ -1,8 +1,5 @@
-"use client";
-
-import { AnchorProvider, Program } from "@coral-xyz/anchor";
-import type { SessionManager } from "@fogo/sessions-idls";
-import { SessionManagerIdl } from "@fogo/sessions-idls";
+import { AnchorProvider } from "@coral-xyz/anchor";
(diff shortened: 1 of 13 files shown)


turbo.json
```diff
@@ -27,15 +27,6 @@
       ],
       "outputs": ["pnpm-lock.yaml"]
     },
-    "build:idl": {
```
cprussin, line 30: I moved this to the package-specific turbo.json because it's very unlikely any other packages will ever implement this

turbo.json
```diff
@@ -117,6 +109,19 @@
     "build:readme": {
       "outputs": ["dist/**/*.md"]
     },
+    "build:types": {
```
cprussin, line 112: I added this one here because it does seem likely to me that other packages may need to implement this one (e.g. maybe we'll decide to use swc to do the build in more places)

apps/sessions-demo/src/components/Home/enable-trading-button.tsx
```diff
@@ -1,8 +1,5 @@
-"use client";
```
guibescos, line 1: why did you drop use-client

guibescos (approved):
great! lgtm

PR #28, opened 2025-07-01
feat: persist and restore sessions in indexeddb

cprussin:
This PR sets up code for persisting & restoring the session in indexeddb.

We are using the [Web Crpyto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) primitives for [CryptoKeyPair](https://developer.mozilla.org/en-US/docs/Web/API/CryptoKeyPair) to persist the session key.  This enables us to mark the private key as non-exportable, and while we can persist it in indexeddb, it is serialized in a way that the secret key material is never directly accessible in javascript.

Doing this was a bit complex as `@solana/web3.js` does not support the Web Crypto API.  However, the successor to `@solana/web3.js`, `@solana/kit`, does support it.  However, we aren't ready to move to `@solana/kit` everywhere, and I think it makes more sense for the external interface of the SDKs to be written in terms of `@solana/web3.js`, which has wildly more adoption.

In order to get around all of this, this PR includes a lot of code to migrate the transaction construction and sending to use `@solana/kit`, while still maintaining the external API in terms of `@solana/web3.js` interfaces.  This requires a bit of conversion back and forth, and a lot of use of `@solana/compat`


packages/sessions-sdk/src/index.ts
```diff
@@ -248,7 +272,8 @@ type EstablishSessionResult = ReturnType<
 >;
 
 export type Session = {
-  sessionKey: Keypair;
+  sessionKeyAddress: Address;
```
guibescos, line 275: seems inconsistent to store this as `Address` instead of `PublicKey`

packages/sessions-sdk/src/index.ts
```diff
@@ -48,32 +52,51 @@ export const establishSession = async (
...
-  const [intentInstruction, startSessionInstruction] = await Promise.all([
-    buildIntentInstruction(options, sessionKey, tokenInfo),
-    buildStartSessionInstruction(options, sessionKey, tokenInfo),
-  ]);
-  const result = await options.adapter.sendTransaction(sessionKey, [
+  const [intentInstruction, startSessionInstruction, sessionKeypair] =
+    await Promise.all([
+      buildIntentInstruction(options, sessionKey, tokenInfo),
+      buildStartSessionInstruction(options, sessionKey, tokenInfo),
+      fromLegacyKeypair(sessionKey, false /* extractable */),
```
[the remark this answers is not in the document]
cprussin: ah yes it should be possible now since we only use the public key for the legacy transaction instructions from web3.js, but the actual signing is done by @solana/kit.  I'll update this.

packages/sessions-sdk/src/index.ts
cprussin: Ah good call.

Frankly this is a bit clunky in general -- we expose the publickey in two separate formats.  I am considering just exporting a function that takes the session and returns the legacy `PublicKey` rather than making it a field

PR #30, opened 2025-07-02
fix: fix handling of decimal token limits

cprussin:


programs/session-manager/src/intents/message.rs
```diff
@@ -4,6 +4,7 @@ use crate::{
 };
 use anchor_lang::prelude::*;
 use chrono::DateTime;
+use rust_decimal::{dec, Decimal};
```
guibescos, line 7: this import is only used in tests, can you move it to the test module?

cprussin: yep, good call, will fix

guibescos (approved):
lgtm

PR #32, opened 2025-07-02
chore: use checked versions of token program instructions

guibescos:
The unchecked versions are deprecated according to some sources

### DIFF
--- a/apps/sessions-demo/package.json
+++ b/apps/sessions-demo/package.json
@@ -39,6 +39,7 @@
     "next": "catalog:",
     "react": "catalog:",
     "react-dom": "catalog:",
+    "swr": "catalog:",
     "tailwind-merge": "catalog:",
     "tailwindcss-animate": "catalog:",
     "zod": "catalog:"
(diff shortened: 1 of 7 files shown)


apps/sessions-demo/src/components/Home/use-trade.ts
```diff
@@ -39,10 +39,11 @@ export const useTrade = (
         new AnchorProvider(connection, {} as Wallet, {}),
       ).methods
         .exampleTransfer(new BN(100))
-        .accounts({
+        .accountsPartial({
           sessionKey: session.sessionPublicKey,
           sink: sinkAta,
           userTokenAccount: userTokenAccount,
+          mint: NATIVE_MINT,
```
cprussin, line 46: Don't merge this yet, I just updated https://github.com/fogo-foundation/fogo-sessions/pull/29 so that `mint` is now an argument of this hook so you'll want to rebase that when I merge once CI passes

cprussin (approved):

PR #33, opened 2025-07-02
chore: fetch chain id from the network

guibescos:


apps/sessions-demo/src/components/Home/use-airdrop.ts
```diff
@@ -56,7 +56,7 @@ export const useAirdrop = (
     );
 
     return result;
-  }, [session, appendTransaction, amount, connection]);
+  }, [session, appendTransaction, amount, connection, mint]);
```
cprussin, line 59: thanks for fixing this

apps/sessions-demo/src/components/Home/use-session.ts
```diff
@@ -160,6 +163,24 @@ const restoreSession = async (
     storedSession.sessionKey,
   );
 
+const fetchChainId = async (connection: Connection) => {
```
cprussin, line 166: shouldn't we do this is `@fogo/sessions-sdk` when constructing the message automatically?  Is there ever a use case for using a chain id for a chain that doesn't match the one you're connected to?

cprussin: Makes sense, when I move things to the react sdk I'll also set up a context provider which handles fetching this & the ALT

PR #37, opened 2025-07-04
Feat/allow tokens as mints

guibescos:


packages/sessions-sdk/src/index.ts
```diff
@@ -100,6 +100,24 @@ const createSession = async (
...
+  Symbol = "Symbol",
+  Mint = "Mint",
+}
+
+const SymbolOrMint = {
+  Symbol: (symbol: string) =>
+    ({
+      type: SymbolOrMintType.Symbol,
+      symbol,
+    }) as const,
```
cprussin, line 113: Minor nit but it's better to declare `as const` on the smallest scope needed since you usually don't want types to be narrowed more than they usually are, so in this case you'd do `Symbol: (symbol: string) => ({ type: SymbolOrMintType.Symbol as const, symbol })`

packages/sessions-sdk/src/index.ts
```diff
@@ -100,6 +100,24 @@ const createSession = async (
...
+  Mint: (mint: PublicKey) =>
+    ({
+      type: SymbolOrMintType.Mint,
+      mint,
+    }) as const,
+};
```
cprussin, line 119: I'm not necessarily recommending you change this, but in this case, you may be able to get by without using a tagged union.  Since `PublicKey` and `string` are not overlapping types, you can just change the type of `symbolOrMint` below to `PublicKey | string` and disambiguate between them by using `value instanceof PublicKey`.

Generally I reach for the tagged union pattern here when either I have values with types that are overlapping but which I need to disambiguate, or states that may not have associated data that I need to indicate, or simply values which are fairly complex where disambiguation without the tag would be clunky and obnoxious (for instance, if it's a much nicer api for the consumer to be able to `switch` on the different tags vs looking for existence of specific fields).

In cases like this where you just have two possible types and they're easy to disambiguate, a simple union is fine and saves some boilerplate.

cprussin (approved):

PR #49, opened 2025-07-09
fix: use env domain in demo app

guibescos:


cprussin (approved):
ah sorry for breaking this

PR #50, opened 2025-07-09
feat: dynamically allocate sessions

guibescos:
Before this change, all sessions on-chain accounts were created with fixed size of 200.
Since sessions have many variable size fields, it was easy to exceed this limit.
After this change, we instead compute the size of the on-chain representation of the session and allocate the right amount of data.


cprussin (approved):
makes sense but you might want to have someone who knows rust / solana better than I do take a look

PR #65, opened 2025-07-19
chore: extend session duration

cprussin:
Until we add controls for setting the session expiration and prompts for when the session has expire, let's just set the default session duration to a long period so people don't get tripped up by this in testnet.


guibescos (approved):
tyvm

PR #81, opened 2025-07-22
feat: add burn instructions

guibescos:


cprussin (approved):
lol not sure i understand this at all but I guess it makes sense

PR #84, opened 2025-07-23
feat: make paymaster config optional & use new public paymaster

cprussin:


pnpm-lock.yaml
```diff
@@ -163,10 +163,10 @@ catalogs:
       specifier: ^19.0.0
       version: 19.1.0
     react-aria:
-      specifier: 3.41.1
+      specifier: ^3.41.1
       version: 3.41.1
     react-aria-components:
-      specifier: 1.10.1
+      specifier: ^1.10.1
```
guibescos, line 169: this seems from another PR

cprussin: Yeah, I forgot to check it in with a prior PR, I'm surprised CI didn't catch it....

PR #91, opened 2025-07-29
Paymaster should simulate tx and reject if fee is too high

jayantk:


services/paymaster/src/api.rs
```diff
@@ -59,6 +62,57 @@ pub fn validate_transaction(
...
+    if let Some(err) = simulation_result.value.err {
+        return Err((
+            StatusCode::INTERNAL_SERVER_ERROR,
+            format!("Transaction simulation failed: {err:?}"),
+        ));
+    }
+
+    // Check if the sponsor account balance change exceeds the maximum permissible value
+    if let Some(accounts) = simulation_result.value.accounts {
+        if let Some(Some(account)) = accounts.first() {
```
cprussin, line 96: what's with the double wrapping?

cprussin (approved):

services/paymaster/src/api.rs
jayantk: accounts is `Vec<Option<Account>>` (defined by the rpc) and so first returns `Option<Option<Account>>`.

PR #95, opened 2025-07-30
fix: only show wallet popup when first connecting wallet

cprussin:
This avoids showing the wallet popup when restoring the session, which has been very common feedback.  It makes sense to only show it when first establishing the session, where it's relevant context.

### DIFF
--- a/packages/sessions-sdk-react/src/session-provider.tsx
+++ b/packages/sessions-sdk-react/src/session-provider.tsx
@@ -474,7 +474,6 @@ const getNextState = (
       }
       case StateType.CheckingStoredSession:
       case StateType.Established:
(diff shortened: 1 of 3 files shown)


packages/sessions-sdk-react/src/session-provider.tsx
```diff
@@ -474,7 +474,6 @@ const getNextState = (
       }
       case StateType.CheckingStoredSession:
       case StateType.Established:
-      case StateType.RestoringSession:
```
cprussin, line 477: This state had been replaced by `CheckingStoredSession` and I guess I forgot to remove the dead code

jayantk (approved):

PR #117, opened 2025-08-18
feat: support send & receive

cprussin:
requires https://github.com/fogo-foundation/token/pull/3 to land in the token program


programs/intent-transfer/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+            self.decimals,
+        )
+    }
+}
+
+impl<'info> TryFrom<&mut SendTokens<'info>> for VerifiedSendTokens<'info> {
+    type Error = Error;
+
+    fn try_from(accounts: &mut SendTokens<'info>) -> Result<Self> {
+        let intent: Intent<Message> = accounts.sysvar_instructions.as_ref().try_into()?;
```
guibescos, line 90: seems like this intent is replayable :(

guibescos (commented):
I think we need to figure out replayability

packages/solana-intents/src/version.rs
```diff
@@ -0,0 +1,140 @@
...
+                input: input.to_string(),
+            })),
+            Err(Err::Failure(Error { code, input })) => Err(Err::Failure(Error {
+                code,
+                input: input.to_string(),
+            })),
+            Err(Err::Incomplete(needed)) => Err(Err::Incomplete(needed)),
+        }
+    }
+}
```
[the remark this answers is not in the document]
cprussin: I love it, way cleaner.  Will do this

packages/sessions-sdk-react/src/toast.module.scss
```diff
@@ -0,0 +1,46 @@
...
+
+.toastRegion {
+  position: fixed;
+  top: theme.spacing(4);
+  right: theme.spacing(12);
+  display: flex;
+  flex-flow: column nowrap;
+  gap: theme.spacing(2);
+  outline: none;
+  z-index: 9999999;
```
alexcambose, line 11: nit, do we have some sort of layered system for these?

packages/sessions-sdk-react/src/toast.module.scss
```diff
@@ -0,0 +1,46 @@
+@use "./theme";
+
+.toastRegion {
+  position: fixed;
```
cprussin, line 4: Note to self: need to wrap this in `@include theme.reset`

packages/sessions-sdk-ts/src/index.ts
```diff
@@ -489,3 +498,102 @@ export type Session = {
...
+    ),
+    await buildTransferIntentInstruction(program, options, symbol),
+    await program.methods
+      .sendTokens()
+      .accounts({
+        destination: destinationAta,
+        mint: options.mint,
+        source: sourceAta,
+        // eslint-disable-next-line unicorn/no-null
+        metadata: null,
```
guibescos, line 548: is it expected that metadata appears twice in this object?

packages/sessions-sdk-ts/src/index.ts
```diff
@@ -489,3 +498,102 @@ export type Session = {
...
+    getMint(options.adapter.connection, options.mint),
+  ]);
+  const message = new TextEncoder().encode(
+    [
+      TRANSFER_MESSAGE_HEADER,
+      serializeKV({
+        version: "0.1",
+        chain_id: options.adapter.chainId,
+        token: symbol ?? options.mint.toBase58(),
+        amount: amountToString(options.amount, decimals),
```
guibescos, line 574: it seems strange that the user provides this as a string then you convert to bigint and back to string here. Should you just use the string that the user has given you (after checking?)

packages/sessions-sdk-ts/src/index.ts
```diff
@@ -489,3 +498,102 @@ export type Session = {
...
+) => {
+  const [nonce, { decimals }] = await Promise.all([
+    getNonce(program, options.walletPublicKey),
+    getMint(options.adapter.connection, options.mint),
+  ]);
+  const message = new TextEncoder().encode(
+    [
+      TRANSFER_MESSAGE_HEADER,
+      serializeKV({
+        version: "0.1",
```
guibescos, line 571:
```suggestion
        version: `${CURRENT_MAJOR}.${CURRENT_MINOR}`,
```

packages/sessions-sdk-rs/Cargo.toml
```diff
@@ -15,7 +15,7 @@ anchor = ["anchor-lang"]
 anchor-idl = ["anchor-lang/idl-build"]
 
 [dependencies]
-borsh = { version = ">=0.10.0, <2.0.0"}
+borsh = { version = ">=0.10.0, <1.0.0"}
```
guibescos, line 18: reconsider this, new versions of solana program want borsh 1.x.x

packages/sessions-sdk-react/src/field.tsx
```diff
@@ -0,0 +1,88 @@
...
+}) => (
+  <TextFieldImpl className={clsx(styles.textField, className)} {...props}>
+    <Field label={label} description={description}>
+      <Input className={styles.input ?? ""} />
+      {controls && <div className={styles.controls}>{controls}</div>}
+    </Field>
+  </TextFieldImpl>
+);
+
+export const NumberField = ({
```
guibescos, line 35: seems unused

programs/intent-transfer/src/error.rs
```diff
@@ -0,0 +1,55 @@
...
+            }
+            IntentError::IncorrectInstructionId => IntentTransferError::IncorrectInstructionId,
+            IntentError::SignatureVerificationUnexpectedHeader => {
+                IntentTransferError::SignatureVerificationUnexpectedHeader
+            }
+            IntentError::ParseFailedError(_) => IntentTransferError::ParseFailedError,
+            IntentError::DeserializeFailedError(_) => IntentTransferError::DeserializeFailedError,
+        }
+    }
+}
```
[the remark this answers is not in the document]
guibescos: no 😞

packages/sessions-sdk-ts/src/index.ts
cprussin: lol no, I guess I had this as `null` initially but forgot to remove it when I did the real implementation.  Good catch, will fix.

packages/sessions-sdk-ts/src/index.ts
cprussin: Yeah, I need to revisit this, the original idea was that `@fogo/session-sdk` (the package that's platform agnostic and can be used from any js code) would always take `bigint`, because there's not really a valid assumption that the value is even coming from user input.  However, that means everywhere we take `bigint`, we also have to resolve the mint to figure out the number of decimals and convert to string.

I'm starting to think it makes more sense to just take strings everywhere in the sdk.  I don't even think we need to check the string, it will fail in the contract if the string is invalid, and the React sdk has form validation to prevent passing an invalid value down.

packages/sessions-sdk-rs/Cargo.toml
cprussin: Ah so the reason I set this down was that when I installed some packages, something updated borsh to 1.x.x, but the install failed due to some breaking changes in borsh.   I'd need to revisit to recall exactly what the issue was but I believe that something in the sessions sdk is actually incompatible with borsh 1.x.x

packages/sessions-sdk-react/src/field.tsx
cprussin: I originally planned to use this but then ended up doing something else, however I decided to keep the code because it will be tree-shaken out and it will likely be a useful component down the line

packages/sessions-sdk-react/src/toast.module.scss
alexcambose: yeah I think that would be nice to see everything (including this outlier) in the mapping

PR #118, opened 2025-08-19
chore: separate faucet and paymaster

guibescos:
I'm doing this change because the paymaster being the wFOGO faucet too was interefering with the ability of the paymaster to receive wFOGO and unwrap it.


apps/sessions-demo/src/config/server.ts
```diff
@@ -40,3 +42,19 @@ export const RPC = envWithDefaults("RPC", {
...
+  localnet:
+    "[156,172,157,152,254,131,214,144,112,176,48,254,231,131,231,245,253,166,29,34,62,108,57,5,90,112,56,190,248,162,193,216,37,30,112,184,13,184,180,63,215,19,244,249,200,64,62,242,252,197,243,232,246,17,16,154,188,86,89,208,205,56,234,17]",
+  testnet: undefined,
+});
+
+if (faucetKey === undefined) {
+  throw new Error("The environment variable FAUCET_KEY is required.");
+}
+
+export const FAUCET_SIGNER = await createKeyPairSignerFromBytes(
```
guibescos, line 58: feels bad to await here

PR #119, opened 2025-08-20
feat: tollbooth

guibescos:
This PR introduces the tollbooth program, a program that caps the spend of the paymaster in each sponsored transaction.

For transactions to be sponsored by the paymaster they need to add a `tollboth::Enter` instruction as their first instruction and `tollboth::Exit` as their last instruction.

In a first stage, we can use it to remove transaction simulation from the paymaster server, reducing latency.
It also enables a fully sustainable paymaster by setting max_sponsor_spending to 0. We will work to make this mode easier for apps in a future PR.

### DIFF
--- a/packages/sessions-idls/package.json
+++ b/packages/sessions-idls/package.json
@@ -27,7 +27,7 @@
     "build:esm": "swc src -d dist/esm --strip-leading-paths --copy-files -C jsc.experimental.keepImportAttributes=true && echo '{\"type\":\"module\"}' > dist/esm/package.json",
     "build:package-json": "transform-package-json ./package.json dist/package.json",
     "build:types": "tsc --project tsconfig.build.json",
-    "build:idl": "mkdir -p src/idl && mkdir -p src/types && anchor idl build --program-name session-manager --out-ts src/types/session-manager.ts --out src/idl/session-manager.json && anchor idl build --program-name example --out-ts src/types/example.ts --out src/idl/example.json && anchor idl build --program-name chain-id --out-ts src/types/chain-id.ts --out src/idl/chain-id.json && anchor idl build --program-name domain-registry --out-ts src/types/domain-registry.ts --out src/idl/domain-registry.json",
+    "build:idl": "mkdir -p src/idl && mkdir -p src/types && anchor idl build --program-name session-manager --out-ts src/types/session-manager.ts --out src/idl/session-manager.json && anchor idl build --program-name example --out-ts src/types/example.ts --out src/idl/example.json && anchor idl build --program-name chain-id --out-ts src/types/chain-id.ts --out src/idl/chain-id.json && anchor idl build --program-name domain-registry --out-ts src/types/domain-registry.ts --out src/idl/domain-registry.json && anchor idl build --program-name tollbooth --out-ts src/types/tollbooth.ts --out src/idl/tollbooth.json",
     "fix:format": "prettier --write .",
     "fix:lint": "eslint --fix . --max-warnings 0",
     "test:format": "run-jest --colors --selectProjects format",
--- a/packages/sessions-sdk-ts/src/adapter.ts
+++ b/packages/sessions-sdk-ts/src/adapter.ts
@@ -1,6 +1,6 @@
 import type { Wallet } from "@coral-xyz/anchor";
 import { AnchorProvider } from "@coral-xyz/anchor";
-import { ChainIdProgram } from "@fogo/sessions-idls";
+import { ChainIdProgram, TollboothProgram } from "@fogo/sessions-idls";
 import {
   fromLegacyPublicKey,
   fromLegacyTransactionInstruction,
@@ -91,6 +91,7 @@ export const createSolanaWalletAdapter = async (
     options.addressLookupTableAddress,
   );
   const sponsor = await getSponsor(options);
+  const paymasterConfig = await getPaymasterConfig(options);
   return {
     connection: options.connection,
     payer: sponsor,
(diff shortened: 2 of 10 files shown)


services/paymaster/src/api.rs
```diff
@@ -51,73 +55,186 @@ pub async fn validate_transaction(
...
+                    return Err((
+                        StatusCode::BAD_REQUEST,
+                        "Tollbooth instructions must be the first and last instruction only"
+                            .to_string(),
+                    ));
+                }
+            }
+        }
+    } else {
+        // Simulate the transaction to check sponsor SOL spending
```
guibescos, line 185: keep this for backward compatibility

programs/tollbooth/src/lib.rs
```diff
@@ -0,0 +1,61 @@
+#![allow(unexpected_cfgs)] // warning: unexpected `cfg` condition value: `anchor-debug`
+
+use anchor_lang::prelude::*;
+
+declare_id!("7czSB69NDz3QJSveAga7igHR86ZkESEou9q845xzPkgf");
```
guibescos, line 5: todo: grind a nice key

packages/sessions-idls/package.json
```diff
@@ -27,7 +27,7 @@
...
     "build:types": "tsc --project tsconfig.build.json",
-    "build:idl": "mkdir -p src/idl && mkdir -p src/types && anchor idl build --program-name session-manager --out-ts src/types/session-manager.ts --out src/idl/session-manager.json && anchor idl build --program-name example --out-ts src/types/example.ts --out src/idl/example.json && anchor idl build --program-name chain-id --out-ts src/types/chain-id.ts --out src/idl/chain-id.json && anchor idl build --program-name domain-registry --out-ts src/types/domain-registry.ts --out src/idl/domain-registry.json",
+    "build:idl": "mkdir -p src/idl && mkdir -p src/types && anchor idl build --program-name session-manager --out-ts src/types/session-manager.ts --out src/idl/session-manager.json && anchor idl build --program-name example --out-ts src/types/example.ts --out src/idl/example.json && anchor idl build --program-name chain-id --out-ts src/types/chain-id.ts --out src/idl/chain-id.json && anchor idl build --program-name domain-registry --out-ts src/types/domain-registry.ts --out src/idl/domain-registry.json && anchor idl build --program-name tollbooth --out-ts src/types/tollbooth.ts --out src/idl/tollbooth.json",
```
cprussin, line 30: We should split this up into separate tasks or something, this is getting a bit absurd lol

packages/sessions-sdk-ts/src/adapter.ts
```diff
@@ -175,6 +183,16 @@ const getSponsor = async (
...
 };
 
+const getPaymasterConfig = async (
+  options: Parameters<typeof createSolanaWalletAdapter>[0],
+) => {
+  const paymasterUrl = "paymaster" in options ? options.paymaster : DEFAULT_PAYMASTER;
+    const response = await fetch(
+      new URL("/api/config", paymasterUrl),
+    );
+    return z.object({ max_sponsor_spending: z.number() }).transform(({ max_sponsor_spending }) => ({ maxSponsorSpending: max_sponsor_spending })).parse(await response.json());
```
cprussin, line 193: Need to run `turbo fix`

programs/tollbooth/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub fn exit<'info>(
+        ctx: Context<'_, '_, '_, 'info, Exit<'info>>,
+        max_allowed_spending: u32,
+    ) -> Result<()> {
+        require_gte!(
+            ctx.accounts
+                .sponsor
+                .lamports()
+                .saturating_add(u64::from(max_allowed_spending)),
+            ctx.accounts.sponsor_snapshot.lamports,
```
cprussin, line 29: Super minor nit but I don't know about you but reading this seemed backwards to me, I think it's more obvious to compare `max_allowed_spending` to the difference between the value at enter and the current value than it is to do it this way

programs/tollbooth/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+cpi = ["no-entrypoint"]
+no-entrypoint = []
+no-idl = []
+no-log-ix-name = []
+idl-build = ["anchor-lang/idl-build", "anchor-spl/idl-build"]
+
+[dependencies]
+anchor-lang = { workspace = true }
+anchor-spl = { workspace = true }
+fogo-sessions-sdk = { path = "../../packages/sessions-sdk-rs" }
```
cprussin, line 21: I don't think the sessions sdk is used

services/paymaster/src/api.rs
```diff
@@ -51,73 +55,186 @@ pub async fn validate_transaction(
...
+        if let Some(accounts) = simulation_result.value.accounts {
+            if let Some(Some(account)) = accounts.first() {
+                let current_balance = account.lamports;
+                // We need to get the pre-transaction balance to calculate the change
+                let pre_balance = rpc.get_balance(sponsor).map_err(|err| {
+                    (
+                        StatusCode::INTERNAL_SERVER_ERROR,
+                        format!("Failed to get sponsor balance: {err}"),
+                    )
+                })?;
```
cprussin, line 226: Shouldn't this be done before simulating in case other txs come through while simulating that change the balance?

services/paymaster/src/api.rs
```diff
@@ -51,73 +55,186 @@ pub async fn validate_transaction(
...
+            ));
+        }
+        Ok(())
+    })?;
+
+    // If the transaction is using the tollbooth, we don't need to simulate it
+    if instructions
+        .first()
+        .map(|i| i.program_id(transaction.message.static_account_keys()) == &TOLLBOOTH_PROGRAM_ID)
+        .unwrap_or_default()
```
cprussin, line 74: The code in this conditional I think can be cleaned up so you don't have to make the expectations and can more clearly enumerate the cases.  Something like this:

```rs
let first_instruction_program_id = instruction.first().map(|i| i.program_id(transaction.message.static_account_keys());
let last_instruction_program_id = instructions.last().map(|i|i.program_id(transaction.message.static_account_keys());

match (first_instruction_program_id, last_instruction_program_id) {
  (Some(&TOLLBOOTH_PROGRAM_ID), Some(&TOLLBOOTH_PROGRAM_ID)) => ...,
  ...
}
```

cprussin (approved):

PR #120, opened 2025-08-22
fix: handle object errors

guibescos:
I noticed while debugging some on-chain errors like the ones with type `TransactionResultType` are sometimes objects.
I think it'd be better not to return Unknown Error for these. Let me know what you think.


cprussin:
@guibescos please make sure to run `pnpm changeset`

PR #126, opened 2025-08-27
feat: confirm txs in the paymaster

guibescos:


services/paymaster/src/rpc.rs
```diff
@@ -0,0 +1,74 @@
...
+    rpc: &RpcClient,
+    transaction: &VersionedTransaction,
+    config: RpcSendTransactionConfig,
+) -> Result<ConfirmationResult, ErrorResponse> {
+    let recent_blockhash = transaction.message.recent_blockhash();
+    let signature = rpc
+        .send_transaction_with_config(transaction, config)
+        .map_err(to_error_response)?;
+
+    for status_retry in 0..GET_STATUS_RETRIES {
```
cprussin, line 43: Is it possible to stream?

cprussin (approved):

services/paymaster/src/rpc.rs
guibescos: Yes it is, we might want to revisit it later.

PR #130, opened 2025-09-03
feat: read on-chain whitelist

guibescos:
I want to avoid config duplication by having the paymaster read the on-chain domain records and use them as a baseline for program whitelists.
(This might change once we have more complicated tx filtering in the paymaster)

### DIFF
--- a/services/paymaster/src/api.rs
+++ b/services/paymaster/src/api.rs
@@ -326,7 +326,7 @@ async fn get_on_chain_program_whitelists(
                 .get_domain_record_address()
         })
         .collect::<Vec<_>>();
-    
+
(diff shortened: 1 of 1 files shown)


cprussin (commented):
The code change makes sense and looks good, but I'm not convinced this is the right functionality.

For one, I could see cases where a team wants to enable sessions to interact with a given program but doesn't want to authorize the paymaster to sign for it.  It seems unlikely, but reasonably possible that it will be needed.

For two, we will have sync issues when the onchain state changes.

I suggest holding back on this change until we have a better idea of what the filtering logic will end up looking like.

cprussin (approved):

PR #131, opened 2025-09-03
feat: add paymaster query string to typescript sdks

guibescos:


packages/sessions-sdk-ts/src/adapter.ts
```diff
@@ -186,12 +188,16 @@ const buildTransaction = async (
...
   options: Parameters<typeof createSolanaWalletAdapter>[0],
+  domain: string,
 ) => {
   if ("sponsor" in options) {
     return options.sponsor;
   } else {
     const response = await fetch(
-      new URL("/api/sponsor_pubkey", options.paymaster ?? DEFAULT_PAYMASTER),
+      new URL(
+        `/api/sponsor_pubkey?domain=${domain}`,
```
cprussin, line 198: You should set the domain query param using `searchParams.set` instead so you don't have subtle issues caused by url escaping:

```ts
const url = new URL("/api/sponsor_pubkey", options.paymaster ?? DEFAULT_PAYMASTER);
url.searchParams.set("domain", domain);
const response = await fetch(url);
```

packages/sessions-sdk-ts/src/adapter.ts
```diff
@@ -218,13 +224,14 @@ const sponsorAndSendResponseSchema = z
...
   transaction: Transaction,
+  domain: string,
 ): Promise<TransactionResult> => {
   if ("sendToPaymaster" in options) {
     return options.sendToPaymaster(transaction);
   } else {
     const response = await fetch(
       new URL(
-        "/api/sponsor_and_send?confirm=true",
+        `/api/sponsor_and_send?confirm=true&domain=${domain}`,
```
cprussin, line 234: Same comment, this should use `searchParams.set

cprussin (changes requested):

packages/sessions-sdk-ts/src/adapter.ts
guibescos: done

cprussin (approved):

PR #135, opened 2025-09-04
fix: replace react-timeago with a custom component

cprussin:
The `react-timeago` library does not ship esm correctly and it's caused issues for the typing, and some of the developers using the sdk have reported that they can't use the latest release due to issues with their environment interacting with this library.

It's a pretty simple component.  Rather than fighting with the library to get it to work correctly, I figure it's better to just home-roll this.

### DIFF
--- a/packages/sessions-sdk-react/src/session-button.tsx
+++ b/packages/sessions-sdk-react/src/session-button.tsx
@@ -12,7 +12,7 @@ import { XCircleIcon } from "@phosphor-icons/react/dist/ssr/XCircle";
 import { PublicKey } from "@solana/web3.js";
 import { Scanner } from "@yudiel/react-qr-scanner";
 import { QRCodeSVG } from "qrcode.react";
-import type { ComponentProps, FormEvent, ReactNode, JSX } from "react";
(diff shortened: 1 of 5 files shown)


packages/sessions-sdk-react/src/session-button.tsx
```diff
@@ -57,17 +55,12 @@ import {
   useTokenAccountData,
 } from "./use-token-account-data.js";
 
+const ONE_SECOND_IN_MS = 1000;
+const ONE_MINUTE_IN_MS = 60 * ONE_SECOND_IN_MS;
+const ONE_HOUR_IN_MS = 60 * ONE_MINUTE_IN_MS;
+const ONE_DAY_IN_MS = 24 * ONE_HOUR_IN_MS;
```
guibescos, line 61: how do you feel about these being duplicated in session-limits.tsx

guibescos (approved):

packages/sessions-sdk-react/src/session-button.tsx
cprussin: I use these all the time lol, I never really felt it was worth centralizing them and exporting them from somewhere since they're just so trivial

PR #138, opened 2025-09-05
revert: fogo sessions domain default

guibescos:
This broke the tilt setup

### DIFF
--- a/apps/sessions-demo/src/config/server.ts
+++ b/apps/sessions-demo/src/config/server.ts
@@ -41,7 +41,7 @@ export const RPC = envWithDefaults("RPC", {
 
 export const FOGO_SESSIONS_DOMAIN = envWithDefaults("FOGO_SESSIONS_DOMAIN", {
   localnet: undefined,
-  testnet: "https://sessions-example.fogo.io"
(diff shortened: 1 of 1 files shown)


apps/sessions-demo/src/config/server.ts
```diff
@@ -39,8 +39,7 @@ export const RPC = envWithDefaults("RPC", {
   testnet: "https://testnet.fogo.io",
 });
 
-export const FOGO_SESSIONS_DOMAIN =
-  process.env.FOGO_SESSIONS_DOMAIN ?? "https://sessions-example.fogo.io";
+export const FOGO_SESSIONS_DOMAIN = process.env.FOGO_SESSIONS_DOMAIN;
```
cprussin, line 42:
```suggestion
export const FOGO_SESSIONS_DOMAIN = envWithDefaults("FOGO_SESSIONS_DOMAIN", {
  localnet: undefined,
  testnet: "https://sessions-example.fogo.io"
});
```

Otherwise the testnet paymaster rejects everything...

cprussin (changes requested):

PR #140, opened 2025-09-05
fix: fix z-indexing issues

cprussin:
Some teams are still reporting issues with z-indexing in some places.  The issue is that react-aria sometimes sets a z-index on some of the overlay elements, but it's inconsistent and often not a high enough z-index.

The fix is to override react-aria's z-indexing using our layering system.  It's unfortunate that we have to use `!important`, but react-aria sets the z-index on the element directly so we have no other choice.


alexcambose (approved):
nice, looks like it's the best approach to solve this

PR #142, opened 2025-09-05
feat: paymaster constraints

anihamde:
This PR adds a set of constraints for the paymaster that can be used to filter transactions to fund.


services/paymaster/src/api.rs
```diff
@@ -88,16 +89,30 @@ pub async fn validate_transaction(
...
             Ok(())
         })?;
 
+    let mut matches_variation = false;
+    for variation in tx_variations {
+        if validate_transaction_against_variation(transaction, variation, sponsor).is_ok() {
+            matches_variation = true;
+            break;
+        }
+    }
```
cprussin, line 108: minor nit but imo cleaner as

```rs
let matches_variation = tx_variations.iter().any(|variation|
  validate_transaction_against_variation(transaction, variation, sponsor).is_ok()
)
```

services/paymaster/src/api.rs
```diff
@@ -157,6 +172,231 @@ pub async fn validate_transaction(
     Ok(())
 }
 
+pub fn validate_transaction_against_variation(
+    transaction: &VersionedTransaction,
+    tx_variation: &TransactionVariation,
+    sponsor: &Pubkey,
+) -> Result<(), (StatusCode, String)> {
+    match tx_variation {
+        TransactionVariation::V1(variation) => validate_transaction_against_variation_v1(transaction, variation, sponsor),
```
cprussin, line 181: nice

services/paymaster/src/api.rs
```diff
@@ -157,6 +172,231 @@ pub async fn validate_transaction(
...
+            return Err((
+                StatusCode::BAD_REQUEST,
+                format!(
+                    "Transaction instruction program ID {} does not match expected ID {} for variation {}",
+                    program_id,
+                    constraint.program,
+                    variation.name
+                ),
+            ));
+        }
```
cprussin, line 216: Is there any reason we might want to collect all the errors rather than just returning the first we encounter?

services/paymaster/src/api.rs
```diff
@@ -157,6 +172,231 @@ pub async fn validate_transaction(
...
+
+        for data_constraint in &constraint.data {
+            check_data_constraint(&instr.data, data_constraint)?;
+        }
+    }
+
+    Ok(())
+}
+
+pub fn check_account_constraint(
```
cprussin, line 244: Should we think about making this function able to represent either a limit to the number of accounts, or a filter accepting exactly a given list of accounts and no others?

services/paymaster/src/api.rs
```diff
@@ -157,6 +172,231 @@ pub async fn validate_transaction(
...
+                    StatusCode::BAD_REQUEST,
+                    format!(
+                        "Transaction instruction missing account at index {} for variation {}",
+                        i, variation.name
+                    ),
+                )
+            })?;
+            let account = transaction.message.static_account_keys()[*account_index as usize];
+            // TODO: do fix this...
+            let user_dummy = &Pubkey::new_unique();
```
cprussin, line 231: I still don't think this will be possible without hitting the rpc

services/paymaster/src/api.rs
```diff
@@ -157,6 +172,231 @@ pub async fn validate_transaction(
...
+            ContextualPubkey::Explicit{pubkey: pk} => pk,
+            ContextualPubkey::User => &user,
+            ContextualPubkey::Session => &session,
+            ContextualPubkey::Sponsor => &sponsor,
+        };
+        if account == *pk_to_check {
+            passes_include = true;
+            break;
+        }
+    }
```
cprussin, line 300: again minor but imo cleaner with `Iter::any`

services/paymaster/src/constraint.rs
```diff
@@ -0,0 +1,148 @@
...
+	pub max_gas_spend: u64,
+}
+
+
+#[derive(Serialize, Deserialize, ToSchema)]
+pub struct RateLimits {
+	user_per_min: u64,
+	session_per_min: u64,
+	ip_per_min: u64,
+}
```
cprussin, line 28: would be nice if we could track and enforce spending caps on these too, but idk if it's feasible

services/paymaster/src/constraint.rs
```diff
@@ -0,0 +1,148 @@
...
+
+pub fn compare_primitive_data_types(a: PrimitiveDataType, constraint: &DataConstraintSpecification) -> Result<(), String> {    
+    let meets = match constraint {
+        DataConstraintSpecification::LessThan(value) => match (a, value) {
+            (PrimitiveDataType::U8(Some(a)), PrimitiveDataType::U8(Some(b))) => a < *b,
+            (PrimitiveDataType::U16(Some(a)), PrimitiveDataType::U16(Some(b))) => a < *b,
+            (PrimitiveDataType::U32(Some(a)), PrimitiveDataType::U32(Some(b))) => a < *b,
+            (PrimitiveDataType::U64(Some(a)), PrimitiveDataType::U64(Some(b))) => a < *b,
+            (PrimitiveDataType::Bool(Some(_)), PrimitiveDataType::Bool(Some(_))) => false, // LessThan not applicable for bool
```
cprussin, line 95: I think this should be an error

services/paymaster/src/constraint.rs
```diff
@@ -0,0 +1,148 @@
...
+            }
+            return Ok(());
+        },
+    };
+
+    if meets {
+        Ok(())
+    } else {
+        Err("Constraint not met".into())
+    }
```
cprussin, line 147: is this needed?  just return `meets` directly

services/paymaster/src/constraint.rs
```diff
@@ -0,0 +1,244 @@
...
+        pubkey: Pubkey,
+    },
+    Sponsor,
+    Signer {
+        index: i8,
+    },
+    UserDelegator,
+}
+
+pub trait ContextualPubkeyTrait {
```
cprussin, line 78: Minor nit but I don't think this trait name tells me much about what the trait actually does

cprussin (approved):

services/paymaster/src/utils.rs
```diff
@@ -0,0 +1,26 @@
...
+where
+    D: Deserializer<'de>,
+{
+    let sol_value: f64 = f64::deserialize(deserializer)?;
+
+    // Convert SOL to lamports (1 SOL = 1,000,000,000 lamports)
+    let lamports = (sol_value * 1_000_000_000.0) as u64;
+
+    Ok(lamports)
+}
```
guibescos, line 26: `serde.rs`?

services/paymaster/src/api.rs
```diff
@@ -157,6 +153,288 @@ pub async fn validate_transaction(
...
+            if constr.required {
+                return result;
+            } else {
+                constraint = constraint_iter.next();
+            }
+        } else {
+            instruction = instruction_iter.next();
+            instruction_index += 1;
+
+            constraint = constraint_iter.next();
```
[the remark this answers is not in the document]
anihamde: you could. i don't think it makes a huge difference? it's clear as written to me rn

services/paymaster/src/constraint.rs
```diff
@@ -0,0 +1,236 @@
...
+
+#[derive(Serialize, Deserialize, ToSchema)]
+pub enum DataConstraintSpecification {
+    LessThan(PrimitiveDataValue),
+    GreaterThan(PrimitiveDataValue),
+    EqualTo(Vec<PrimitiveDataValue>),
+    Neq(Vec<PrimitiveDataValue>),
+}
+
+pub fn compare_primitive_data_types(
```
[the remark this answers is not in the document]
anihamde: would probably have to use generics, which would involve more impl blocks and probably more complexity than is worth

PR #143, opened 2025-09-06
chore: improve paymaster errors

guibescos:


cprussin:
@guibescos need to run `pnpm changeset`

PR #145, opened 2025-09-09
feat: add limits exceeded error to sessions sdk

cprussin:

PR #150, opened 2025-09-10
feat: gracefully handle trading past limits

cprussin:


packages/sessions-sdk-react/src/session-provider.tsx
```diff
@@ -71,6 +70,7 @@ const ONE_DAY_IN_MS = 24 * ONE_HOUR_IN_MS;
 const DEFAULT_SESSION_DURATION = 14 * ONE_DAY_IN_MS;
 
 const ERROR_CODE_SESSION_EXPIRED = 4_000_000_000;
+const ERROR_CODE_SESSION_LIMITS_EXCEEDED = 4_000_000_008;
```
anihamde, line 73: where do you get these numbers from? why are non-consecutive?

packages/sessions-sdk-react/src/session-provider.tsx
```diff
@@ -214,6 +214,20 @@ const SessionProvider = ({
...
+      <ModalDialog
+        heading="This trade exceeds your set limits"
+        message="Would you like to increase your session limits?"
+        isOpen={sessionState.type === StateType.RequestingIncreasedLimits}
+        onOpenChange={onExtendSessionOpenChange}
+      >
+        {isEstablished(sessionState) && (
+          <RenewSessionsContents
+            sessionState={sessionState}
+            enableUnlimited={enableUnlimited}
```
anihamde, line 226: if the trade exceeds set limits, then should this be `false`? as the user has already created a limited session?

anihamde (commented):
to be sure, this doesn't account for the case where the user has exactly used up to their limits, right?

packages/sessions-sdk-react/src/session-provider.tsx
cprussin: no this option means the app has requested unlimited sessions to be an available option to the user, not that the user has selected it

packages/sessions-sdk-react/src/session-provider.tsx
cprussin: this error code comes from https://github.com/fogo-foundation/fogo-sessions/pull/145/files#diff-b8c5c4505cf04822f63a57243414d1ac40d4aaad09773d593ba632f026c458d9R37 which I already had to merge so I could release an upgraded sdk in order to use that error in the token program.  It's non-consecutive because there are other enum members for `SessionError` that we don't handle in the frontend, e.g. if there's an `InvalidAccountVersion` we don't do anything to handle it in the sdk.

cprussin:
> to be sure, this doesn't account for the case where the user has exactly used up to their limits, right?

@anihamde no it does, what will happen is when you trade exactly up to your limits, the token program sets the delegate to `COption::None`.  The next request would then trigger https://github.com/fogo-foundation/token/blob/main/program/src/processor.rs#L318-L320 and return the `LimitsExceeded` error, which is caught by the same UI code path here

PR #166, opened 2025-09-16
fix: improve signer constraints

anihamde:
This PR improves the signer constraints given that signer order (besides the fee payer) is not easily derivable when there are more than 2 non-fee-paying signers.

### DIFF
--- a/services/paymaster/src/constraint.rs
+++ b/services/paymaster/src/constraint.rs
@@ -53,7 +53,6 @@ impl VariationProgramWhitelist {
 pub struct VariationOrderedInstructionConstraints {
     pub name: String,
     pub instructions: Vec<InstructionConstraint>,
-    pub rate_limits: RateLimits,
     pub max_gas_spend: u64,
 }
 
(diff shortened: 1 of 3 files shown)


services/paymaster/src/constraint.rs
```diff
@@ -293,37 +285,33 @@ impl ContextualPubkey {
                 }
             }
 
-            ContextualPubkey::Signer { index } => {
-                let index_uint = if *index >= 0 {
-                    *index as usize
-                } else if (-*index as usize) <= signers.len() {
-                    signers.len() - (-*index as usize)
+            ContextualPubkey::FeePayer => {
+                if expect_include == (account == signers.first().expect("No signatures in transaction")) {
```
cprussin, line 289: The fee payer should always be the paymaster, I don't see any reason to make this configurable

cprussin (changes requested):
I think we should get rid of `FeePayer` and hardcode a check that ensures the paymaster account is the fee payer

services/paymaster/src/constraint.rs
anihamde: Good catch, `FeePayer` is actually a duplicate of the `Sponsor` variant. I'll remove

anihamde: We already have a check that ensures the paymaster account is the fee payer [here](https://github.com/fogo-foundation/fogo-sessions/blob/a96383b3eed0b562675b5668aa32b5004e9f4314/services/paymaster/src/api.rs#L140-L149)

cprussin (approved):

PR #168, opened 2025-09-16
fix: add serde default tags

anihamde:
This PR adds serde default tags to the paymaster constraint API, so as to default relevant unspecified fields to empty vectors.

### DIFF
--- a/services/paymaster/src/constraint.rs
+++ b/services/paymaster/src/constraint.rs
@@ -52,6 +52,7 @@ impl VariationProgramWhitelist {
 #[derive(Serialize, Deserialize)]
 pub struct VariationOrderedInstructionConstraints {
     pub name: String,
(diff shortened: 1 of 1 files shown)


services/paymaster/src/constraint.rs
```diff
@@ -211,7 +214,9 @@ impl InstructionConstraint {
 #[derive(Serialize, Deserialize, Clone)]
 pub struct AccountConstraint {
     pub index: u16,
+    #[serde(default)]
     pub include: Vec<ContextualPubkey>,
```
guibescos, line 218: nit: I don't see the usecase of providing an AccountConstraint with empty include and exclude

anihamde: no such use case. just want to avoid having to specify empty arrays whenever possible.

PR #170, opened 2025-09-17
feat: add basic metrics to paymaster

anihamde:
This PR adds some basic metrics to the paymaster service. In particular, this PR adds the following metrics:

- Transactions validated (count)
- Transactions submitted to RPC (count)
- Gas spent (count)
- Gas spent (histogram)

using the following labels:

- App domain name
- Transaction variation name
- Result (of validation/submission)

The metrics are published at the 4000 port of the service.

TODOs:

- [ ] Add relevant labels (signers?)
- [ ] Add more metrics (paymaster balance)

### DIFF
--- a/services/paymaster/src/metrics.rs
+++ b/services/paymaster/src/metrics.rs
@@ -1,0 +1,27 @@
+use axum_prometheus::metrics;
+
+pub const TRANSACTION_VALIDATION_COUNT: &str = "paymaster_transaction_validation_total";
+pub fn obs_validation(domain: String, variation: String, result_validation: String) {
+    let labels = &[("domain", domain), ("variation", variation), ("result", result_validation)];
+    metrics::counter!(TRANSACTION_VALIDATION_COUNT, labels).increment(1);
+}
+
+pub const TRANSACTION_SEND_COUNT: &str = "paymaster_transaction_send_total";
+pub fn obs_send(domain: String, variation: String, result_confirmation: Option<String>) {
+    let mut labels = vec![("domain", domain), ("variation", variation)];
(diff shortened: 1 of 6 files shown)


cprussin (dismissed):

services/paymaster/src/metrics.rs
```diff
@@ -0,0 +1,46 @@
...
+    variation: String,
+    result_confirmation: Option<String>,
+    lamports: u64,
+) {
+    let mut labels = vec![("domain", domain), ("variation", variation)];
+    if let Some(result) = result_confirmation {
+        labels.push(("result", result));
+    }
+    metrics::counter!(GAS_SPEND_COUNT, &labels).increment(lamports);
+    metrics::histogram!(GAS_SPEND_HISTOGRAM, &labels).record(lamports as f64);
```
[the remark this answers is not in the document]
anihamde: you're totally right, histogram gives us count as well. the `as` is probably fine in practice here, but i did something a bit more explicit to be safe

services/paymaster/src/rpc.rs
```diff
@@ -21,6 +21,15 @@ pub enum ConfirmationResult {
     },
 }
 
+impl ConfirmationResult {
```
cprussin, line 24: Would it be more idiomatic to make this `Display` or `ToString`?

cprussin (approved):

services/paymaster/src/rpc.rs
anihamde: idk if that's the right use case for those traits since some info is being lost... this representation is just for the label string

PR #172, opened 2025-09-19
chore: address nits

guibescos:


packages/sessions-sdk-ts/src/index.ts
```diff
@@ -400,6 +400,15 @@ const getTokenInfo = async (
 
 type TokenInfo = Awaited<ReturnType<typeof getTokenInfo>>[number];
 
+const serializeU16LE = (value: number) => {
+  return Uint8Array.from([
+    // eslint-disable-next-line unicorn/number-literal-case
+    value & 0xff,
```
anihamde, line 406: maybe i'm just bad at bit shifts and operations, but i think you should comment out what's happening here for clarity

packages/solana-intents/src/lib.rs
```diff
@@ -176,3 +177,55 @@ impl<P> From<borsh::io::Error> for IntentError<P> {
...
+
+    #[test]
+    fn test_raw_message() {
+        let message = b"Fogo Sessions";
+        assert_eq!(
+            Message::deserialize(message).unwrap(),
+            Message::Raw(message.to_vec())
+        );
+
+        let message_long = b"Fogo Sessions Fogo Sessions Fogo Sessions";
```
anihamde, line 225: what's the point of the long messages? just another test case? seems unnecessary

packages/solana-intents/src/lib.rs
```diff
@@ -176,3 +177,55 @@ impl<P> From<borsh::io::Error> for IntentError<P> {
...
+        assert_eq!(
+            Message::deserialize(&message_serialized_long)
+                .unwrap_err()
+                .kind(),
+            std::io::ErrorKind::InvalidData
+        );
+    }
+
+    #[test]
+    fn test_raw_message() {
```
anihamde, line 218: `test_raw_message_roundtrip`?

packages/solana-intents/src/lib.rs
```diff
@@ -176,3 +177,55 @@ impl<P> From<borsh::io::Error> for IntentError<P> {
...
+    fn test_offchain_message_roundtrip() {
+        let offchain_message = OffchainMessage::new(0, "Fogo Sessions".as_bytes()).unwrap();
+
+        assert_eq!(
+            Message::deserialize(&offchain_message.serialize().unwrap()).unwrap(),
+            Message::OffchainMessage(offchain_message.clone())
+        );
+
+        let mut message_serialized_short: Vec<u8> = offchain_message.serialize().unwrap();
+        message_serialized_short.pop();
```
anihamde, line 195: i think these separate test cases, especially when they have different outcomes, should live in different functions

anihamde: doing that would also help to make the test case more readable bc it would be clear why it fails. right now it's not the most readable

packages/sessions-sdk-ts/src/index.ts
cprussin: yes please, 100% agree

cprussin (approved):

PR #174, opened 2025-09-24
feat(sdk): generate sign in token

alexcambose:
* Exposed `sessionKey` instead of `sessionKeyPublic`
* Added generic methods to sign and verify a signature
* Added a method on the react sdk to generate a signed message with the session key


cprussin (changes requested):
Nice direction but some changes that need to happen

cprussin (dismissed):

PR #175, opened 2025-09-25
Fix: remove implicit compute budget

anihamde:
Previously, Compute Budget instructions needed to be specified in v1 transaction variations. However, given the `max_gas_spend` check, this is unnecessary, and apps do not need to specify allowing Compute Budget instructions.

In addition to the `SetComputeUnitLimit` and `SetComputeUnitPrice` instructions, there are also `RequestHeapFrame` and `SetLoadedAccountsDataSizeLimit` instructions (and a deprecated `RequestUnits` instruction). These do not add to the gas cost, and they are not commonly used. Preventing use of these instructions does not appear to be critical at this time; if apps request, we can filter to only permit `SetComputeUnitLimit` and `SetComputeUnitPrice` instructions arbitrarily.

### DIFF
--- a/services/paymaster/src/constraint.rs
+++ b/services/paymaster/src/constraint.rs
@@ -72,6 +72,20 @@ pub struct ContextualDomainKeys {
     pub sponsor: Pubkey,
 }
 
+fn is_compute_budget_instruction(transaction: &VersionedTransaction, instruction_index: usize) -> bool {
+    if let Some(instruction) = transaction.message.instructions().get(instruction_index) {
+        let static_accounts = transaction.message.static_account_keys();
+        let program_id = instruction.program_id(static_accounts);
+        if program_id == &solana_compute_budget_interface::id() {
+            if let Ok(_) = ComputeBudgetInstruction::try_from_slice(&instruction.data) {
+                return true;
+            }
+        }
+    }
+    
+    false
+}
+
 impl VariationOrderedInstructionConstraints {
     pub fn validate_transaction(
         &self,
@@ -80,9 +94,16 @@ impl VariationOrderedInstructionConstraints {
         chain_index: &ChainIndex,
     ) -> Result<(), (StatusCode, String)> {
         let mut instruction_index = 0;
+        let mut constraint_index = 0;
         check_gas_spend(transaction, self.max_gas_spend)?;
 
-        for constraint in self.instructions.iter() {
+        while constraint_index < self.instructions.len() {
+            if is_compute_budget_instruction(transaction, instruction_index) {
+                instruction_index += 1;
+                continue;
+            }
+
+            let constraint = &self.instructions[constraint_index];
             let result = constraint.validate_instruction(
                 transaction,
                 instruction_index,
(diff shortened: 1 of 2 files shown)


services/paymaster/src/constraint.rs
```diff
@@ -95,8 +125,10 @@ impl VariationOrderedInstructionConstraints {
                 if constraint.required {
                     return result;
                 }
+                constraint_index += 1;
             } else {
                 instruction_index += 1;
+                constraint_index += 1;
```
guibescos, line 131: This `constraint_index += 1;` could be outside of the if else

services/paymaster/src/constraint.rs
```diff
@@ -72,6 +72,23 @@ pub struct ContextualDomainKeys {
...
 
+fn is_compute_budget_instruction(
+    transaction: &VersionedTransaction,
+    instruction_index: usize,
+) -> bool {
+    if let Some(instruction) = transaction.message.instructions().get(instruction_index) {
+        let static_accounts = transaction.message.static_account_keys();
+        let program_id = instruction.program_id(static_accounts);
+        if program_id == &solana_compute_budget_interface::id()
+            && ComputeBudgetInstruction::try_from_slice(&instruction.data).is_ok()
```
guibescos, line 83: It bothers me a little bit to have the deserialization here, I feel like it this is not within the responsibility of this while loop which is to check that the transaction conforms to the instruction filters.

I think there's room for throwing errors if:
- there are duplicate compute budget instructions
- there are invalid compute budget instructions
But maybe they belong in the first pass inside `check_gas_spend`?

services/paymaster/src/constraint.rs
anihamde: to me, the explicitness makes it more readable. otherwise, it's more mental effort to process what happens to which indices depending on the result and the `required` bool

services/paymaster/src/constraint.rs
```diff
@@ -72,6 +72,22 @@ pub struct ContextualDomainKeys {
...
+) -> bool {
+    if let Some(instruction) = transaction.message.instructions().get(instruction_index) {
+        let static_accounts = transaction.message.static_account_keys();
+        let program_id = instruction.program_id(static_accounts);
+        if program_id == program_to_match {
+            return true;
+        }
+    }
+
+    false
```
guibescos, line 88: Returning false when the instruction is out of bounds works in the specific logic of this program but unintuitive and could lead to error.

Let's imagine another person decides to do some stuff when `!instruction_matches_program` without looking at the implementation. He would be surprised the programs goes into the if statement when the instruction is out of bounds.

I would make this Option<bool> and you can use unwrap_or() at the callsite?

But the root problem of this issue is it is awkward you have the `transaction.message.instructions().get` in both `instruction_matches_program` and  `validate_instruction`. At this point it might make sense to get the instruction in the while loop and pass the instruction to `instruction_matches_program` and `validate_instruction`, but it's a bigger refactor.

services/paymaster/src/constraint.rs
guibescos: got it

guibescos (approved):

PR #180, opened 2025-09-29
feat: add spans to paymaster

anihamde:
This PR adds spans to trace the timing of transaction validation and submission flows. This allows for keeping track of the relative timings of the flows for different transactions.


services/paymaster/src/rpc.rs
```diff
@@ -39,29 +41,58 @@ fn to_error_response(err: Error) -> ErrorResponse {
...
                 });
             }
             return Err(to_error_response(err));
         }
     };
 
+    confirm_transaction(rpc, &signature, recent_blockhash).await
+}
+
+// Wrapper for send_transaction_with_config to add tracing
```
anihamde, line 70: i don't love the extremely shallow interface here, but it was done to keep all the tracing in the macros. we could also get rid of this and just directly trace in the non-confirmation path with

```
tracing::info_span!("send_transaction", tx_hash = %transaction.signatures[0])
```

but i slightly preferred separating out all the tracing into macros. open to feedback

services/paymaster/src/api.rs
```diff
@@ -358,22 +368,21 @@ async fn sponsor_and_send_handler(
...
-        let signature = state
-            .chain_index
-            .rpc
-            .send_transaction_with_config(
-                &transaction,
-                RpcSendTransactionConfig {
-                    skip_preflight: true,
-                    ..RpcSendTransactionConfig::default()
-                },
+        let signature = send_transaction(
```
guibescos, line 371: fyi: we are trying to get rid of this branch. New versions of the typescript sdk use the version where the paymaster waits for confirmation

guibescos (approved):

guibescos (approved):

PR #185, opened 2025-10-01
feat: optimize some components for small screens

cprussin:
This commit adds a `compact` session button variant which is designed for small screens.  It also optimizes the toast and the session popover widget for small / mobile screens

### DIFF
--- a/.changeset/hip-seas-sort.md
+++ b/.changeset/hip-seas-sort.md
@@ -1,0 +1,5 @@
+---
+"@fogo/sessions-sdk-react": patch
+---
+
+Optimize components for mobile & add compact main button variant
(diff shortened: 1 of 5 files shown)


anihamde (approved):
generally kinda feel that the expansion of the wallet UI near screen bottom is a bit weird, but could be just me. will defer to you on that.

cprussin:
> generally kinda feel that the expansion of the wallet UI near screen bottom is a bit weird, but could be just me. will defer to you on that.

I think it's OK, this is a fairly standard UX layout for popovers on mobile sized screens; it's a bit weird if you're testing it in webinspector or something but it makes more sense on a phone screen

PR #186, opened 2025-10-01
Update getAccountInfo call to use 'confirmed' commitment

koneko-chann:
To avoid the race condition when the previous is not confirmed


cprussin (approved):
Thanks!

cprussin:
@koneko-chann please fix the CI failure before I merge, you can run `pnpm fix` here

guibescos (approved):

koneko-chann:
Sorry for late reply, i got the fever and havent checked the mail, thanks for the merge

PR #192, opened 2025-10-01
fix: paymaster audit feedback

anihamde:
This PR addresses two issues mentioned in the audit:

1. lack of check on the domain PDA in `start_session_instruction_constraint`
2. `compute_gas_spent` not accounting for precompile signatures


services/paymaster/src/constraint.rs
```diff
@@ -730,9 +731,32 @@ pub fn process_compute_budget_instructions(
     Ok(priority_fee)
 }
 
+/// The Solana precompile programs that verify signatures.
+pub const PRECOMPILE_SIGNATURE_PROGRAMS: &[Pubkey] = &[
+    ed25519_program::ID,
+    secp256k1_program::ID,
+    secp256r1_program::ID,
```
anihamde, line 738: Fogo does not have this precompile live on testnet, but

1. any transactions invoking this address as a program should fail
2. the current calculation would overestimate the gas spend which is the safer path (sans checking whether the precompile is active)

guibescos (approved):
thanks!

PR #197, opened 2025-10-02
feat: add button to copy mint address

cprussin:


cprussin:
> i like the way the mint appears! one comment is that when i click on most of the body of the token rectangle, i expect to be taken to a page that allows me to send or receive this token, but this is only the case if i click the blue send icon. maybe a TODO for post-mainnet, though it may be trivial to add

yes we already have design proposals for this page, it's coming but yes post mainnet.  Definitely not trivial (and having a page that only has a send button is going to look awful so we have to wait until we have time implement stuff that people would expect on a token details page like a price graph etc)

PR #200, opened 2025-10-03
fix: parser inconsistencies

guibescos:
Some bugs that I found while working on extra keys

### DIFF
--- a/packages/solana-intents/src/key_value.rs
+++ b/packages/solana-intents/src/key_value.rs
@@ -2,7 +2,7 @@ use nom::{
     branch::alt,
     bytes::complete::{tag, take_till1, take_while1},
     character::complete::{alphanumeric1, anychar, char, line_ending},
-    combinator::{eof, map, map_opt, opt, peek, recognize},
+    combinator::{eof, map, map_opt, not, opt, peek, recognize},
     error::ParseError,
     multi::many_till,
     sequence::{delimited, preceded, separated_pair},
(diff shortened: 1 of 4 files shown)


packages/solana-intents/src/key_value.rs
```diff
@@ -55,14 +55,20 @@ where
             alt((
                 delimited(
                     tag(" "),
-                    take_till1(|c: <I as Input>::Item| c.is_newline()),
+                    take_till1(|c: <I as Input>::Item| c.is_newline() || c.as_char() == '\r'),
```
guibescos, line 58: this was inconsistent with line_ending, with checks both` /r/n` and `/n`

programs/intent-transfer/src/message.rs
```diff
@@ -59,6 +59,7 @@ where
                 tag_key_value("recipient"),
                 tag_key_value("nonce"),
             ),
+            eof,
```
guibescos, line 62: disallow adding extra data at the end of the intent

packages/solana-intents/src/key_value.rs
cprussin: I think we can use `not_line_ending` here instead of this line, e.g.

```rs
delimited(tag(" "), not_line_ending, alt((line_ending, eof)))
```

cprussin (approved):
Nice catches!

packages/solana-intents/src/key_value.rs
guibescos: https://github.com/fogo-foundation/fogo-sessions/pull/204

PR #201, opened 2025-10-03
fix: check session expiration in verifylogintoken

guibescos:


cprussin (approved):

packages/sessions-sdk-ts/src/index.ts
```diff
@@ -857,5 +857,15 @@ export const verifyLogInToken = async (
...
+    connection,
+    new PublicKey(payload.sessionPublicKey),
+  );
+  if (!sessionAccount) return;
+
+  if (sessionAccount.expiration.getTime() < Date.now()) {
+    throw new Error("The session associated with this login token has expired");
+  }
+
+  return sessionAccount;
```
cprussin, line 870: This can be a bit more concise:

```suggestion
  if (sessionAccount?.expiration.getTime() < Date.now()) {
    throw new Error("The session associated with this login token has expired");
  } else {
    return sessionAccount;
  }
```

I personally dislike early returns and try to avoid them, but of course you could leave out the `else` if you wanted.

alexcambose:
Thanks @guibescos!

alexcambose (approved):

packages/sessions-sdk-ts/src/index.ts
guibescos: mmmm I since this function is using many early returns already I'm going to keep it like this.
Also when reading your suggestion i had to think a little bit on what happens if sessionAccount is undefined, which makes me think it reads worse if you're not deep into typescript.

PR #224, opened 2025-10-10
feat(paymaster): Database Loader

alexcambose:
# 📋 Database Paymaster Configuration
This PR adds the ability for the paymasster to consume its config from a Database, while keeping the existing logic for loading it from a TOML file (to be deprecated soon).
As an addition to this change, we also support configuring the paymaster via `.env` files.

## Added fIle structure
* `/src/config_manager` - this contains the logic to **load** the config from the **DB** and **Filesystem** (TOML) and also to **spawn a task** that will **refetch** and **update the config** used by the paymaster asynchronously.
* `/src/db` - this has the **database connection pool**, **migrations**, logic for seeding the database with an existing configuration object (in our case the one that was generated reading the TOML file, this will be deleted after we fully migrate to DB only) and also the actual query that reads all the data from the db. 

## Key points
* we **mix** relational database structure with `jsonb` fields for storing the variations
* every 10 seconds we **refetch** all the configurations in the DB and internally reassign it to be used for validation. This ensures that we keep the paymaster endpoints lean and fast. We might want to use something like https://www.postgresql.org/docs/current/sql-notify.html to further **optimize** this, especially when we'll have the Admin UI ready.
* 

## Database structure
Untitled

### Dependencies before merging
* we need a production & development database, Neon DB with PgSql 18 should be good I think
* we need to change the config in the deployments and add the database url


services/paymaster/src/config_manager/load_config.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+/// Compare the two configs to make sure they are the same.
+#[allow(dead_code)]
+fn compare_configs(config1: &Config, config2: &Config) -> bool {
+    if config1.domains.len() != config2.domains.len() {
+        return false;
+    }
+    for domain1 in &config1.domains {
```
anihamde: better rust practice is to use [iters](https://doc.rust-lang.org/book/ch13-02-iterators.html) rather than loops. for example, you could write this as:

```
config1.domains.iter().all(|domain1| {
    config2.domains.iter().find(|domain2| {
        domain1 == domain2 // or whatever equality check function you want here
    }).is_some()
})
```

services/paymaster/src/config_manager/load_config.rs
```diff
@@ -0,0 +1,141 @@
...
+                    return false;
+                }
+            }
+        }
+    }
+
+    true
+}
+
+pub async fn load_config(config_path: &str) -> Result<Config> {
```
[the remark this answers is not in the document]
anihamde: you could make it so the filepath and the db url are command line args, and you can [only specify one](https://docs.rs/clap/latest/clap/struct.ArgGroup.html#method.conflicts_with) of them

.env.example
```diff
@@ -0,0 +1,6 @@
+DATABASE_URL=
+OTEL_EXPORTER_OTLP_ENDPOINT=
+MNEMONIC_FILE=
```
anihamde: some of these were set to cli args. i think that's slightly better practice bc you don't have to hardcode these env variables whenever you want to make a change

services/paymaster/Cargo.toml
```diff
@@ -51,13 +53,18 @@ solana-transaction = {workspace = true}
...
 tower-http = {version = "0.6.6", features = ["cors"]}
 tracing = "0.1.41"
 tracing-opentelemetry = "0.31.0"
 tracing-subscriber = {version = "0.3.20", features = ["env-filter", "json"]}
 utoipa = "5.4.0"
 utoipa-axum = "0.2.0"
+url = "2.5.7"
+rand = "0.9.2"
+uuid = {version = "1.18.1", features = ["v4"]}
+serde_json = "1.0.145"
```
anihamde: you can get rid of this too

services/paymaster/src/main.rs
```diff
@@ -80,18 +100,24 @@ async fn main() -> anyhow::Result<()> {
...
+    // ----- load initial config -----
+    let config = config_manager::load_config::load_config(&cli.config).await?;
+    let mnemonic =
+        std::fs::read_to_string(&cli.mnemonic_file).expect("Failed to read mnemonic_file");
+    let domains: SharedDomains = Arc::new(RwLock::new(api::get_domain_state_map(
+        config.domains,
+        &mnemonic,
+    )));
+
+    // ----- spawn config refresher -----
```
anihamde: can get rid of this comment

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,177 @@
...
+    Ok(Config {
+        domains: map.into_values().collect(),
+    })
+}
+
+/// Hash a plaintext password using Argon2.
+fn hash_password(plain: &str) -> String {
+    let salt = SaltString::generate(&mut OsRng);
+
+    // Argon2 with default params (Argon2id v19)
```
anihamde: can get rid of these comments

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,177 @@
...
+            parts[parts.len() - 1]
+        ))
+    } else {
+        Some(host.to_string())
+    }
+}
+/// Insert a user(email, password) into the database.
+async fn insert_user(domain_url: &Url) -> Result<Uuid, sqlx::Error> {
+    let password = hash_password("admin");
+    let email = format!("admin@{}", &registrable_domain(domain_url).unwrap());
```
anihamde: you generally don't want to use `unwrap` in rust because it will end up panicking. That will kill the entire thread instead of handling the error gracefully, which you can do by something like this:

```
let email = format!("admin@{}", &registrable_domain(domain_url).ok_or(...))
```

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,177 @@
...
+          v.instructions  AS instructions
+        FROM domain_config d
+        JOIN variation v ON v.domain_config_id = d.id
+        ORDER BY d.id
+        "#,
+    )
+    .fetch_all(pool::pool())
+    .await?;
+
+    let mut map: HashMap<Uuid, Domain> = HashMap::new();
```
anihamde: instead of making a mut map and adding to it via for loop, try a more functional approach a la:

```
let map: HashMap<Uuid, Domain> = rows.into_iter().fold(HashMap::new(), |mut acc, r| {
    acc.entry(r.domain_id)
        .or_insert_with(|| Domain {
            domain: r.domain.clone(),
            enable_session_management: r.enable_session_management,
            enable_preflight_simulation: r.enable_preflight_simulation,
            tx_variations: Vec::new(),
        })
        .tx_variations
        .push(r.instructions.0);
    acc
});
```

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,177 @@
...
+    .bind(name)
+    .fetch_one(pool())
+    .await?;
+
+    Ok(app.0)
+}
+
+async fn insert_domain_config(app_id: &Uuid, domain: &Domain) -> Result<Uuid, sqlx::Error> {
+    let domain_config = sqlx::query_as::<_, (Uuid,)>(
+        "INSERT INTO domain_config (app_id, domain, enable_session_management, enable_preflight_simulation) VALUES ($1, $2, $3, $4) RETURNING id",
```
anihamde: if you run sqlx::prepare, it should generate a cached query statement which optimizes running this query with different params.

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,177 @@
...
+    .bind(domain_config_id)
+    .bind(Json(&variation))
+    .fetch_one(pool())
+    .await?;
+
+    Ok(variation.0)
+}
+
+/// Seed the database from the config.
+pub async fn seed_from_config(config: &Config) -> Result<(), sqlx::Error> {
```
anihamde: the server shouldn't be making any edits to the db in run mode. it should purely be reading from the db. another reason to not try loading both ways

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,177 @@
...
+fn hash_password(plain: &str) -> String {
+    let salt = SaltString::generate(&mut OsRng);
+
+    // Argon2 with default params (Argon2id v19)
+    let argon2 = Argon2::default();
+
+    // Hash password to PHC string ($argon2id$v=19$...)
+    argon2
+        .hash_password(plain.as_bytes(), &salt)
+        .unwrap()
```
anihamde: see note about unwrap(), better practice is to return a Result for `hash_password`

services/paymaster/src/db/pool.rs
```diff
@@ -0,0 +1,27 @@
+use sqlx::{postgres::PgPoolOptions, Pool, Postgres};
+use tokio::sync::OnceCell;
+
+static POOL: OnceCell<sqlx::PgPool> = OnceCell::const_new();
+
+/// Initialize the database connection pool and runs any pending migrations.
+pub async fn init_db_connection(database_url: String) -> Result<(), sqlx::Error> {
```
anihamde: you could pass in a `&str` here since you don't need to own it in the implementation of this function. better practice generally to pass in references

anihamde (commented):
Overall good job on the first Rust PR! I would recommend trying to simplify what you're going for here. I think you can get rid of the dual loading and not try to edit the db from within this run mode (should only read from the db). you can add a different mode to run migrations.

services/paymaster/src/config_manager/load_config.rs
alexcambose: ahh right! I'm going to delete this but will keep the iters thing in mind!

.env.example
alexcambose: Not sure I understand this.

I feel like as the number of env variables is growing it makes sense to have a .env file to store these, otherwise locally running the paymaster would mean always setting a bunch of cli args and I would be cumbersome.

services/paymaster/Cargo.toml
```diff
@@ -8,16 +8,18 @@ repository = "https://github.com/fogo-foundation/fogo-sessions"
...
+argon2 = "0.5.3"
 axum = "0.8.4"
 axum-extra = {version = "0.10.1", features = ["typed-header"]}
 axum-prometheus = "0.9.0"
 base64 = "0.22.1"
 bincode = "1.3.3"
 borsh = {workspace = true}
 bytemuck = {workspace = true}
-clap ={version = "4.5.41", features = ["derive"]}
+clap ={version = "4.5.41", features = ["derive","env"]}
```
guibescos:
```suggestion
clap = {version = "4.5.41", features = ["derive","env"]}
```

services/paymaster/src/main.rs
```diff
@@ -1,43 +1,64 @@
-use crate::config::load_config;
+use std::sync::Arc;
+
+extern crate dotenv;
```
guibescos: I dont think you need `extern crate dotenv;`

services/paymaster/src/main.rs
```diff
@@ -1,43 +1,64 @@
-use crate::config::load_config;
+use std::sync::Arc;
```
guibescos: why is this import statement separated from the others

services/paymaster/src/main.rs
```diff
@@ -1,43 +1,64 @@
...
 mod metrics;
 mod rpc;
 mod serde;
 
-#[derive(Parser)]
+type DomainStateMap = std::collections::HashMap<String, api::DomainState>;
+type SharedDomains = Arc<RwLock<DomainStateMap>>;
+
+#[derive(Debug, Parser)]
+#[clap(author, version, about, long_about = None)]
```
guibescos: I feel like `author` and `long_about` can be removed since the crate has no `author` in the manifest and `long_about = None` is a noop in this context

services/paymaster/src/constraint.rs
```diff
@@ -9,9 +9,10 @@ use solana_pubkey::Pubkey;
 use solana_sdk_ids::{ed25519_program, secp256k1_program, secp256r1_program};
 use solana_transaction::versioned::VersionedTransaction;
 
-use crate::{rpc::ChainIndex, serde::deserialize_pubkey_vec};
+use crate::rpc::ChainIndex;
+use crate::serde::{deserialize_pubkey_vec, serialize_pubkey_vec};
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone, Debug)]
```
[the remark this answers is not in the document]
guibescos: I dont mind Debug since maybe it improves developer experience by making it easier to print these objects.
Clone is more problematic since it can encourage unnecessary cloning

services/paymaster/src/db/pool.rs
```diff
@@ -0,0 +1,27 @@
...
+        .await?;
+
+    // run migrations
+    sqlx::migrate!("src/db/migrations").run(pool).await?;
+
+    Ok(())
+}
+
+/// Get the database connection pool.
+pub fn pool() -> &'static Pool<Postgres> {
```
guibescos:
```suggestion
pub fn pool() -> &'static PgPool {
```

services/paymaster/src/config_manager/load_config.rs
```diff
@@ -0,0 +1,141 @@
...
+    db::config::seed_from_config(&file_config).await?;
+
+    let mut db_config = load_db_config().await?;
+
+    assign_config_defaults(&mut db_config);
+    assign_config_defaults(&mut file_config);
+
+    // this was used when testing to make sure we have the same config structure from the database as the one from the toml file
+    // if !compare_configs(&db_config, &file_config) {
+    //     // throw error
```
guibescos: I don't like commiting dead code even though it was useful during testing

services/paymaster/src/db/config.rs
alexcambose: hmm tried something like this: 
```
 let email = format!(
        "admin@{}",
        registrable_domain(domain_url).ok_or(anyhow!("invalid domain in URL: {}", domain_url))
    );
```
but it returns:`admin@Ok("invariant.app")`

services/paymaster/src/db/pool.rs
alexcambose: okay got it, thanks!

services/paymaster/src/api.rs
```diff
@@ -208,9 +208,10 @@ async fn sponsor_and_send_handler(
     Query(SponsorAndSendQuery { domain, .. }): Query<SponsorAndSendQuery>,
     Json(payload): Json<SponsorAndSendPayload>,
 ) -> Result<Json<ConfirmationResult>, ErrorResponse> {
+    let val = state.domains.read().await;
     let domain = get_domain_name(domain, origin)?;
     tracing::Span::current().record("domain", domain.as_str());
-    let domain_state = get_domain_state(&state, &domain)?;
+    let domain_state = get_domain_state(&val, &domain)?;
```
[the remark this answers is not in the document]
anihamde: pass in `state.domains` into `get_domain_state`?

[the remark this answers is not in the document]
alexcambose: yeah, but the issue is here:
```rust
async fn get_domain_state<'a>(
    domains: &'a Arc<RwLock<HashMap<String, DomainState>>>,
    domain_query_parameter: &str,
) -> Result<&'a DomainState, (StatusCode, String)> {
    let domain_state = domains.read().await
        .get(domain_query_parameter)
        .ok_or_else(|| {
            (
                StatusCode::BAD_REQUEST,
                format!(
                    "The domain {domain_query_parameter} is not registered with the paymaster, please either set the domain property in FogoSessionProvider to match your production domain or reach out to the Fogo team to get a paymaster configuration for your app"
                ),
            )
        })?;

    Ok(domain_state)
}


error[E0515]: cannot return value referencing temporary value
   --> services/paymaster/src/api.rs:187:5
    |
176 |     let domain_state = domains.read().await
    |                        -------------------- temporary value created here
...
187 |     Ok(domain_state)
    |     ^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function

```

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,177 @@
...
+        .to_string()
+}
+
+// Get the registrable domain from a URL.
+// Example: https://test.brasa.finance -> brasa.finance
+fn registrable_domain(u: &Url) -> Option<String> {
```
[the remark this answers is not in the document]
alexcambose: later edit: this will be for the username

services/paymaster/migrations/20251020133349_init.up.sql
```diff
@@ -1,7 +1,7 @@
 CREATE TABLE if not exists "user" (
   id uuid PRIMARY KEY default uuidv7(),
-  email text NOT NULL UNIQUE,
-  password text NOT NULL,
+  username text NOT NULL UNIQUE,
+  wallet_address text NOT NULL,
```
alexcambose: later edit: changed from email/password to wallet based auth for the frontend so we'll have username/wallet_address and people can log in with their session key

services/paymaster/src/config_manager/load_config.rs
```diff
@@ -0,0 +1,61 @@
+use arc_swap::ArcSwap;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
+
+use crate::api::{self, DomainState};
+use crate::config_manager::config::Config;
+
+use crate::db;
```
guibescos: I don't like how there are multiple import sections separated by spaces

services/paymaster/migrations/20251020133349_init.up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE if not exists "user" (
+  id uuid PRIMARY KEY default uuidv7(),
+  username text NOT NULL UNIQUE,
```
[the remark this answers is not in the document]
guibescos: I could buy the username is being like an alias for the `wallet_address` like "Ani's wallet"

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,157 @@
...
+#[derive(FromRow)]
+struct Row {
+    domain_id: Uuid,
+    domain: String,
+    enable_session_management: bool,
+    enable_preflight_simulation: bool,
+    instructions: Json<TransactionVariation>,
+}
+
+/// Load the config from the database
```
anihamde: don't need this comment, doesn't add any explanatory value

services/paymaster/src/config_manager/load_config.rs
```diff
@@ -0,0 +1,61 @@
...
+use config::File;
+use tokio::time::interval;
+
+/// Load the config from the file.
+pub fn load_file_config(config_path: &str) -> Result<Config> {
+    let config: Config = config::Config::builder()
+        .add_source(File::with_name(config_path))
+        .build()?
+        .try_deserialize()?;
+    // note we DO NOT want to assign defaults here because this will be used to seed the database and there's no point adding the txs from eg. enable_session_management
```
anihamde: i don't like seeding the db here from file. i am not saying this is strictly wrong; i am just saying i don't think that we should be doing this here in the server. we can host this elsewhere rather than creating code for a one-time operation from the server that we probably won't ever want to run again.

services/paymaster/src/db/pool.rs
```diff
@@ -0,0 +1,31 @@
+use sqlx::{postgres::PgPoolOptions, PgPool};
+use tokio::sync::OnceCell;
+
+static POOL: OnceCell<PgPool> = OnceCell::const_new();
+
+/// Initialize the database connection pool and runs any pending migrations.
```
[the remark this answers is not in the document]
anihamde: although upon immediately reading the function, seems like this is just an outdated comment

services/paymaster/src/api.rs
```diff
@@ -223,11 +223,10 @@ fn get_domain_name(
 }
 
 fn get_domain_state<'a>(
-    state: &'a ServerState,
+    domains_guard: &'a HashMap<String, DomainState>,
```
anihamde: you don't need the `_guard` naming here since the type in the function is just a reference to a hashmap

services/paymaster/Cargo.toml
```diff
@@ -53,13 +54,16 @@ solana-transaction = {workspace = true}
...
+sqlx = { version = "0.8", features = [ "postgres", "runtime-tokio", "tls-native-tls", "uuid" ] }
 tokio = {version = "1.46.1", features = ["full"]}
 tower-http = {version = "0.6.6", features = ["cors"]}
 tracing = "0.1.41"
 tracing-opentelemetry = "0.31.0"
 tracing-subscriber = {version = "0.3.20", features = ["env-filter", "json"]}
 utoipa = "5.4.0"
 utoipa-axum = "0.2.0"
+url = "2.5.7"
+uuid = {version = "1.18.1", features = ["v4"]}
```
anihamde: why do we need `v4` features? i thought we were using v7 above?

services/paymaster/src/config_manager/load_config.rs
```diff
@@ -0,0 +1,61 @@
...
+    // config.assign_defaults();
+    Ok(config)
+}
+
+/// Load the config from the database.
```
anihamde: lo mismo

services/paymaster/src/config_manager/load_config.rs
```diff
@@ -0,0 +1,61 @@
...
+    Ok(config)
+}
+
+/// Spawn a background task to refresh the config every 10 seconds.
+pub fn spawn_config_refresher(
+    mnemonic: String,
+    domains: Arc<ArcSwap<HashMap<String, DomainState>>>,
+) {
+    tokio::spawn(async move {
+        let mut ticker = interval(Duration::from_secs(10));
```
anihamde: nit: make refresh frequency configurable function argument

services/paymaster/src/config_manager/load_config.rs
```diff
@@ -0,0 +1,61 @@
...
+        // First tick fires immediately, we can skip it if we don't want a duplicate load.
+        ticker.tick().await;
+
+        loop {
+            ticker.tick().await;
+            // note that `load_config` also reads from the toml file but that shouldn't be an issue
```
anihamde: this comment seems outdated? `load_db_config` doesn't touch the toml file anymore

anihamde (commented):
looks good! most of my comments are nits, but this made a lot of progress!

services/paymaster/migrations/20251020133349_init.up.sql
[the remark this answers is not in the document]
alexcambose: it's not a requirement for the paymaster, but would be nice to have some sort of human readable identifier for users and for for use who the user is, and we'll use that on the frontend as well.

services/paymaster/src/db/config.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    if let Some(user_result) = existing_user {
+        return Ok(user_result.0);
+    }
+
+    let new_user = sqlx::query_as::<_, (Uuid,)>(
+        "INSERT INTO \"user\" (username, wallet_address) VALUES ($1, $2) RETURNING id",
+    )
+    .bind(&username)
+    .bind(format!("wallet-address-{}", username))
```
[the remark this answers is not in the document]
alexcambose: yes, we don't have the apps/developers wallet addresses so inserting a placeholder for now, not an issue for the paymaster to work but we'll need to replace this with an actual address so they can authenticate in the UI.

services/paymaster/migrations/20251020133349_init.up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE if not exists "user" (
```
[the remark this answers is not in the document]
alexcambose: no, but `user` is a reserved keyword and we need "" as opposed to the rest of the table names

services/paymaster/src/main.rs
```diff
@@ -1,52 +1,34 @@
...
+async fn run_server(opts: cli::RunOptions) -> anyhow::Result<()> {
     let resource = opentelemetry_sdk::Resource::builder()
         .with_attributes(vec![opentelemetry::KeyValue::new(
             "service.name",
             "paymaster-service",
         )])
         .build();
 
-    let otlp_endpoint = cli
+    let otlp_endpoint = opts
```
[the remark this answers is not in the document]
alexcambose: ah yes, also added a `default_value` so `unwrap_or_else` should not even be required anymore

PR #234, opened 2025-10-15
feat: rewrite wallet adapter & session state machine

cprussin:
This commit rewrites the session provider state machine & the wallet adapter code to remove the usage of the `@solana/wallet-adapter-react` contexts.

The Solana wallet adapter library has some major downsides for our use cases:

- it does not expose interfaces that make it easy to hook into state changes, which has led to obscure bugs and nasty code which is hard to understand.

- we cannot easily control the wallet selection UI (this piece could be replaced without replacing the underlying Solana wallet adapter library, but it's much easier to own the UI when we also own the state management).

Rewriting this code to use our own state management drastically simplifies the logic here, resolves some bugs, and enables us to add custom UI for the wallet select flow.  Among other things, this resolves 3 outstanding tickets:

- [UI-136: Allow apps to specify links to TOS agreements and add them to the session init flow](https://linear.app/douro-labs/issue/UI-136/allow-apps-to-specify-links-to-tos-agreements-and-add-them-to-the)
- [UI-263: Fix mobile wallet connect issues](https://linear.app/douro-labs/issue/UI-263/fix-mobile-wallet-connect-issues)
- [UI-267: Fix broken state when restarting tilt](https://linear.app/douro-labs/issue/UI-267/fix-broken-state-when-restarting-tilt)

In this commit I also rename the `adapter.ts` file in the typescript sdk to `context.ts`.  Originally, the intent of that file was to abstract away the parts of the wallet interaction that are specific to Solana.  Over time the structure of code has changed and it's now used to initialize global app context, such as fetching the ALT, the chain ID, and the sponsor key.  So the name `adapter` does not make any sense any more, and it became very obvious how nonsensical it was while I was working on this PR.

### DIFF
--- a/.changeset/purple-pandas-hang.md
+++ b/.changeset/purple-pandas-hang.md
@@ -1,0 +1,5 @@
+---
+"@fogo/sessions-idls": patch
+---
+
+Add the tollbooth program and sessions v0.4
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1623,7 +1623,7 @@ dependencies = [
 
 [[package]]
 name = "fogo-sessions-sdk"
-version = "0.6.1"
+version = "0.7.0"
 dependencies = [
  "anchor-lang",
  "borsh 0.10.4",
@@ -6542,6 +6542,15 @@ dependencies = [
  "futures-sink",
  "pin-project-lite",
  "tokio",
+]
+
+[[package]]
+name = "tollbooth"
+version = "0.1.0"
+dependencies = [
+ "anchor-lang",
+ "anchor-spl",
+ "fogo-sessions-sdk",
 ]
 
 [[package]]
--- a/Tiltfile
+++ b/Tiltfile
@@ -76,7 +76,7 @@ local_resource(
 
 local_resource(
     "paymaster",
-    serve_cmd="cargo run --bin fogo-paymaster",
+    serve_cmd="cargo run --bin fogo-paymaster -- --rpc-url-http http://127.0.0.1:8899 --rpc-url-ws ws://127.0.0.1:8900 --mnemonic-file ./tilt/secrets/mnemonic",
     resource_deps=["svm-localnet"],
 )
 
--- a/packages/sessions-idls/package.json
+++ b/packages/sessions-idls/package.json
@@ -33,6 +33,7 @@
     "build:idl:example": "anchor idl build --program-name example --out-ts src/types/example.ts --out src/idl/example.json",
     "build:idl:intent-transfer": "anchor idl build --program-name intent-transfer --out-ts src/types/intent-transfer.ts --out src/idl/intent-transfer.json",
     "build:idl:session-manager": "anchor idl build --program-name session-manager --out-ts src/types/session-manager.ts --out src/idl/session-manager.json",
+    "build:idl:tollbooth": "anchor idl build --program-name tollbooth --out-ts src/types/tollbooth.ts --out src/idl/tollbooth.json",
     "fix:format": "prettier --write .",
     "fix:lint": "eslint --fix . --max-warnings 0",
     "test:format": "run-jest --colors --selectProjects format",
--- a/packages/sessions-idls/src/index.ts
+++ b/packages/sessions-idls/src/index.ts
@@ -5,11 +5,13 @@ import DomainRegistryIdlImpl from "./idl/domain-registry.json" with { type: "jso
 import ExampleIdlImpl from "./idl/example.json" with { type: "json" };
 import IntentTransferIdlImpl from "./idl/intent-transfer.json" with { type: "json" };
 import SessionManagerIdlImpl from "./idl/session-manager.json" with { type: "json" };
+import TollboothIdlImpl from "./idl/tollbooth.json" with { type: "json" };
 import type { ChainId } from "./types/chain-id.js";
 import type { DomainRegistry } from "./types/domain-registry.js";
 import type { Example } from "./types/example.js";
 import type { IntentTransfer } from "./types/intent-transfer.js";
 import type { SessionManager } from "./types/session-manager.js";
+import type { Tollbooth } from "./types/tollbooth.js";
 
 export type SessionManagerIdl = SessionManager;
 export const SessionManagerIdl = SessionManagerIdlImpl as SessionManagerIdl;
@@ -48,5 +50,13 @@ export const IntentTransferIdl = IntentTransferIdlImpl as IntentTransferIdl;
 export class IntentTransferProgram extends Program<IntentTransferIdl> {
   constructor(provider: AnchorProvider) {
     super(IntentTransferIdl, provider);
+  }
+}
+
+export type TollboothIdl = Tollbooth;
+export const TollboothIdl = TollboothIdlImpl as TollboothIdl;
+export class TollboothProgram extends Program<TollboothIdl> {
+  constructor(provider: AnchorProvider) {
+    super(TollboothIdl, provider);
   }
 }
--- a/packages/sessions-idls/turbo.json
+++ b/packages/sessions-idls/turbo.json
@@ -14,7 +14,8 @@
         "build:idl:example",
         "build:idl:domain-registry",
         "build:idl:intent-transfer",
-        "build:idl:session-manager"
+        "build:idl:session-manager",
+        "build:idl:tollbooth"
       ]
     },
     "build:idl:mkdir": {},
@@ -79,6 +80,17 @@
         "src/idl/session-manager.json",
         "src/types/session-manager.ts"
       ],
+      "dependsOn": ["build:idl:mkdir"]
+    },
+    "build:idl:tollbooth": {
+      "inputs": [
+        "../../programs/tollbooth/src/**/*.rs",
+        "../../programs/tollbooth/Cargo.toml",
+        "../../packages/**/src/**/*.rs",
+        "../../Cargo.toml",
+        "../../Anchor.toml"
+      ],
+      "outputs": ["src/idl/tollbooth.json", "src/types/tollbooth.ts"],
       "dependsOn": ["build:idl:mkdir"]
     },
     "build:types": {
--- a/packages/sessions-sdk-react/src/get-metadata.ts
+++ b/packages/sessions-sdk-react/src/get-metadata.ts
@@ -13,7 +13,7 @@ const metadataSchema = z.record(
   z.string(),
   z.object({
     name: z.string(),
-    symbol: z.string(),
+    symbol: z.string().optional(),
     image: z.string(),
   }),
 );
--- a/packages/sessions-sdk-rs/Cargo.toml
+++ b/packages/sessions-sdk-rs/Cargo.toml
@@ -3,7 +3,7 @@ name = "fogo-sessions-sdk"
 description = "A set of utilities for SVM on-chain programs integrating with Fogo sessions"
 edition = { workspace = true }
 license = { workspace = true }
-version = "0.6.1"
+version = "0.7.0"
 
 [lib]
 crate-type = ["rlib"]
--- a/packages/sessions-sdk-rs/src/lib.rs
+++ b/packages/sessions-sdk-rs/src/lib.rs
@@ -3,6 +3,7 @@ pub mod error;
 pub mod intent_transfer;
 pub mod session;
 pub mod token;
+pub mod tollbooth;
 
 #[cfg(feature = "anchor")]
 const ID: anchor_lang::prelude::Pubkey = session::SESSION_MANAGER_ID;
--- a/packages/sessions-sdk-rs/src/session/mod.rs
+++ b/packages/sessions-sdk-rs/src/session/mod.rs
@@ -1,5 +1,6 @@
 use borsh::BorshSchema;
 use solana_program::account_info::AccountInfo;
+use solana_program::hash::HASH_BYTES;
 use solana_program::pubkey::Pubkey;
 use solana_program::sysvar::clock::Clock;
 use solana_program::sysvar::Sysvar;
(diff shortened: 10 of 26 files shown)


packages/sessions-sdk-react/package.json
```diff
@@ -15,7 +15,7 @@
   "main": "./dist/cjs/index.js",
   "types": "./dist/cjs/index.d.ts",
   "engines": {
-    "node": ">=18"
+    "node": ">=22"
```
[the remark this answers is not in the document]
cprussin: I don't think it's needed:

- all of our users are apps that have been written within the last few months given Fogo is a new chain, there is no good reason why anyone would be on node 20 given no one was even started building Fogo apps until after 22 was already in active LTS
- given we're in a security-sensitive space I think it's reasonable to target active lts as the minimum supported node version
- Moving down to 20 means adding some polyfills / extra null checks for some APIs that we now need that would be executed in a SSR, in particular for some of the code which initializes the mobile wallet adapter.  It's not a big deal, but it's marginally cleaner if we don't have to deal with it

I think it's reasonable in this case to move forward assuming 22 and relax it in a later commit if we really need to

packages/sessions-sdk-ts/src/context.ts
```diff
@@ -80,23 +80,26 @@ export type TransactionResult = ReturnType<
...
-export const createSolanaWalletAdapter = async (
+export const createSessionContext = async (
   options: {
     connection: Connection;
     addressLookupTableAddress?: string | undefined;
     domain?: string | undefined;
   } & (
     | {
         paymaster?: string | URL | undefined;
+        sendToPaymaster?: undefined;
```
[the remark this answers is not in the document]
anihamde: i guess the pattern of checking if something is undefined is better than checking if its variable name is in options?

packages/sessions-sdk-react/src/components/renew-session-modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+  StateType as TokenDataStateType,
+} from "../hooks/use-token-account-data.js";
+import { StateType } from "../session-state.js";
+
+export const RenewSessionModal = () => {
+  const sessionState = useSession();
+
+  const onOpenChange = useCallback(
+    (isOpen: boolean) => {
+      if (!isOpen && "cancel" in sessionState) {
```
anihamde: is there a way to expose `isCanceling` or some other method on `sessionState` to avoid querying for strings?

packages/sessions-sdk-react/src/session-state.ts
```diff
@@ -0,0 +1,120 @@
...
+  endSession: () => void;
+  updateSession: (
+    prevState: StateType,
+    duration: number,
+    limits?: Map<PublicKey, bigint>,
+  ) => void;
+};
+
+export const SessionState = {
+  Initializing: () => ({ type: StateType.Initializing as const }),
```
anihamde: some of these feel extraneous? like do you need `Initializing`, or could that be recast to `UpdatingSession`? and there seems to be some leakage between session state and wallet state, or is that just a naming issue?

packages/sessions-sdk-react/src/components/renew-session-modal.tsx
cprussin: There is but it's entirely redundant, we're not actually querying for strings here, this code is saying "Is the session state in any state that has a `cancel` function?" -- i.e. it would be equivalent but redundant (and thus error prone) to add a flag `canCancel` or have a filter that explicitly enumerates the states that can be canceled.  IMO the implementation here is the simplest -- we cancel the action if the action can be canceled, which is indicated by the existence of a `cancel` method in the state.  If we add or remove a cancel method down the line to a state, the UI will automatically work or prevent the modal from closing.

packages/sessions-sdk-ts/src/context.ts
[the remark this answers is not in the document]
cprussin: In the strictest sense, [optional `?` is not the same as `undefined`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes).  With strict settings, typescript treats an optional as a key that may not exist, where an undefined is a key that may exist with the value `undefined`.

So if I have `{ foo? string }`, then the object may either 1) not have a `foo` key at all, or 2) have a `foo` key with a `string` value.  It may _not_ have a `foo` key with the value `undefined`.

We should generally be using very strict settings for typescript in general, but it's particularly important in libraries, because if a package user uses strict typescript settings, it will cause problems to them if the libraries they want to use do not.  However in this case, we actually don't care if `sendToPaymaster` is not passed at all, or if it's given an `undefined` value.  What this type is basically asserting is that you either optionally pass a `paymaster`, or you pass _both_ a `sendToPaymaster` and a `sponsor`.  If you pass a `paymaster`, you can only pass either of the other two values if the value you pass is `undefined` (and vice versa, if you pass the other two, you're still allowed to pass `paymaster` if the value you pass is `undefined`).

This allows the consumer API to be a bit more flexible and avoid having the end-user having to do nasty spread operations on the props if they had these values determined dynamically.

packages/sessions-sdk-react/src/hooks/use-session.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { PublicKey } from "@solana/web3.js";
+import { createContext, use } from "react";
+
+import type { SessionState } from "../session-state.js";
+
+export const SessionContext = createContext<
+  | {
+      getSessionContext: () => Promise<SessionExecutionContext>;
+      sessionState: SessionState;
+      enableUnlimited: boolean;
```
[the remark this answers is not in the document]
cprussin: `whitelistedTokens` is at this point a fairly poorly named config options -- this actually controls the list of tokens that you can set limits on if you choose to create a limited session.  These two options are actually not connected at all and can be set fully independently -- e.g. you could have any combination of apps that have unlimited sessions enabled/disabled and apps that have tokens available for setting limits -- we actually explicitly handle the case where both are disabled to indicate that you're looking to establish a "view only" session which is useful e.g. if you were using sessions to auth to a database and not an onchain program.

To be honest the mechanism where the app determines which tokens are available for setting limits is a pretty bad UX and needs to be redesigned in a future iteration when we spend more time refining limits, so likely this will go away entirely down the line and the user will instead have tho option to choose any token they want to apply limits to (maybe the app will still get an option to surface specific tokens at the top of the list or something).

packages/sessions-sdk-react/src/components/session-provider.tsx
```diff
@@ -77,841 +73,781 @@ const ERROR_CODE_SESSION_LIMITS_EXCEEDED = 4_000_000_008;
...
+    } else {
+      return ConnectWalletState.RestoredSession(session);
+    }
+  }
 };
 
-const SessionState = {
-  Initializing: () => ({ type: StateType.Initializing as const }),
+enum ConnectWalletStateType {
+  Aborted,
```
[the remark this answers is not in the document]
cprussin: yeah let me add some comments to the code here but to answer your question:

- `Aborted` - the user aborted the connect flow (e.g. by closing the modal) before we finished connecting the wallet
- `Connected` - the user connected the wallet and is now ready to set limits and sign the message to establish the session
- `RestoredSession` - the user connected the wallet and we found a valid session in indexeddb for that wallet, so we can skip the limit setting / message signing part of the flow and go straight to the Established state

packages/sessions-sdk-react/src/session-state.ts
cprussin: > there seems to be some leakage between session state and wallet state

Regarding this, the coupling is very intentional -- the wallet state is in fact a portion of the session state machine.  We could technically break this up into a sub-machine and it might be marginally cleaner to do so, but since this is already and has been part of the public api for a long time I don't intend to change that public api with this PR

packages/sessions-sdk-react/src/components/button.tsx
```diff
@@ -1,27 +1,39 @@
...
-  <ButtonImpl
-    className={clsx(className, styles.button)}
-    data-variant={variant}
-    data-size={size}
-    {...props}
-  />
-);
+  ...otherProps
+}: {
+  className?: Parameters<typeof clsx>[0];
```
alexcambose: curious if there's any type benefit of doing this vs `string`

packages/sessions-sdk-react/src/components/session-provider.tsx
```diff
@@ -77,841 +73,781 @@ const ERROR_CODE_SESSION_LIMITS_EXCEEDED = 4_000_000_008;
...
+            addressSelector: createDefaultAddressSelector(),
+            appIdentity: {
+              uri:
+                // eslint-disable-next-line unicorn/no-typeof-undefined
+                typeof globalThis.window === "undefined"
+                  ? ""
+                  : `${globalThis.window.location.protocol}//${globalThis.window.location.host}`,
+            },
+            authorizationResultCache: createDefaultAuthorizationResultCache(),
+            chain: "mainnet-beta",
```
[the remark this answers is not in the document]
alexcambose: makes sense, I thought this would also me used for sending a tx and the chain id would be used for singing too as part of the tx (eg. EIP-155 in evm)

packages/sessions-sdk-react/src/components/modal-dialog.tsx
```diff
@@ -1,31 +1,49 @@
...
-          <Heading slot="title" className={styles.heading ?? ""}>
-            {heading}
-          </Heading>
-          {message && <div className={styles.message}>{message}</div>}
-          {typeof children === "function" ? children(args) : children}
-        </Dialog>
-      )}
-    </Modal>
-  </ModalOverlay>
+  <AnimatePresence>
```
benduran: similar comment here about being rather generic but potentially useful to have for other projects

packages/sessions-sdk-react/src/components/session-limits-tab.tsx
```diff
@@ -64,7 +64,13 @@ const SessionLimitsForm = ({
           }
           onSubmit={
             "updateSession" in sessionState
```
benduran: it feels like it might be beneficial to store these keys being looked up in sessionState some place. right now, feels a little too much like a magic string that requires a bit of tribal knowledge 😅

packages/sessions-sdk-react/src/components/sign-in-modal.tsx
```diff
@@ -0,0 +1,318 @@
...
+  selectWallet?:
+    | ((wallet: MessageSignerWalletAdapterProps) => void)
+    | undefined;
+  cancel: () => void;
+  termsOfServiceUrl?: string | undefined;
+  privacyPolicyUrl?: string | undefined;
+}) => {
+  const [moreOptionsOpen, setMoreOptionsOpen] = useState(false);
+
+  const installedWallets = useMemo(
```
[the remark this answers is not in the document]
benduran: would only end up needing one `useMemo()`, as well 👍

packages/sessions-sdk-react/src/session-state.ts
```diff
@@ -0,0 +1,120 @@
...
+    type: StateType.SettingLimits as const,
+    ...args,
+  }),
+
+  Established: (args: EstablishedOptions) => ({
+    type: StateType.Established as const,
+    ...args,
+  }),
+
+  UpdatingSession: (
```
benduran: (for the future): this type of state object might be well suited for something battle tested like [zustand](https://www.npmjs.com/package/zustand), which manages the state object outside of React (for performance reasons) but ships with hooks for reading and interacting with it

or if you prefer something a bit more vanilla, [totally-tubular](https://www.npmjs.com/package/totally-tubular)

packages/sessions-sdk-react/src/components/session-limits-tab.tsx
cprussin: This would be equally valid to write this as:

```ts
isEstablished(session) && sessionState.type !== StateType.UpdatingSession
    ? sessionState.updateSession
    : undefined
```

But doing it as I have is less error-prone if state types get added (all we need to do is ensure new states have the callbacks that are relevant and the UI will work itself out) -- plus it's marginally more concise.

IMO doing `"updateSession" in sessionState` is no more of a magic string than checking the state type and then calling `sessionState.updateSession` -- you know you're trying to check if `updateSession` is valid to call on your current state.  In other words, this is not checking an arbitrary string, it's checking for the presence of a discriminator in the discriminated union (and to answer your point about the keys being stored in some place -- they are stored in the union variants of the discriminated union and are easy to see by looking at the `SessionState` type)

cprussin: Another way we could do this is to add a function `canUpdateSession` to `session-state.ts` instead, then it removes the code that does the discrimination from the call site which is probably clearer -- what are your thoughts on that?

packages/sessions-sdk-react/src/components/renew-session-modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+  sessionState: EstablishedSessionState;
+}) => {
+  const state = useTokenAccountData(sessionState);
+  const { enableUnlimited, whitelistedTokens } = useSessionContext();
+
+  switch (state.type) {
+    case TokenDataStateType.Error:
+    case TokenDataStateType.Loaded: {
+      return (
+        <SessionLimits
```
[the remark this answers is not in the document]
cprussin: The problem is that `<SessionLimits />` is used in other contexts, and in particular in the `sign-in-modal.tsx` component when we use it, the state machine can make type-level guarantees that there _will never_ be a loading state (i.e. the information that drives the props we pass to `<SessionLimits />` is not async and so there is no loading state in that call site).

So the async loading & spinner is not in fact a property of `<SessionLimits />` but is rather a property of this usage of it.  That's why the wrapper & state management is here and not in `<SessionLimits />`.

I do think things could be a bit cleaner with `<SessionLimits />` in general, but this part of the UX is due for a significant overhaul (both in the onchain representations & the UX flow of the components), so some added context here is that I don't think it's worth a lot of effort to do much about this since it's all going to be trashed and rewritten soon anyways.

benduran (approved):
LGTM

anihamde (approved):

PR #235, opened 2025-10-16
feat: toolbooth paymaster

guibescos:
This PR adds toll support to the paymaster server.
Next PR we can tackle the UI sdks.

### DIFF
--- a/scripts/src/initialize-chain-id.ts
+++ b/scripts/src/initialize-chain-id.ts
@@ -38,7 +38,7 @@ export const main = async (argv: string[] = hideBin(process.argv)) => {
 
   const url = args.url == "l" ? "http://localhost:8899" : args.url;
 
-  const connection = new Connection(url);
+  const connection = new Connection(url, "confirmed");
   const keypair = Keypair.fromSecretKey(
     Buffer.from(JSON.parse(fs.readFileSync(args.keypair, "utf8"))),
   );


scripts/src/add-program-id-to-domain-registry.ts
```diff
@@ -47,7 +47,7 @@ export const main = async (argv: string[] = hideBin(process.argv)) => {
 
   const url = args.url == "l" ? "http://localhost:8899" : args.url;
 
-  const connection = new Connection(url);
+  const connection = new Connection(url, "confirmed");
```
anihamde: do you want to add this to `initialize-chain-id.ts` as well?

services/paymaster/src/constraint/mod.rs
```diff
@@ -174,19 +183,12 @@ impl InstructionConstraint {
     pub async fn validate_instruction(
         &self,
         transaction: &VersionedTransaction,
+        instruction: &CompiledInstruction,
```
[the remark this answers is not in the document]
anihamde: you could also get rid of instruction_index. i would say we could do that if reextracting the instruction was expensive, but since it isn't, i think it would be better to keep instruction_index for better error messages

services/paymaster/src/constraint/tolls.rs
```diff
@@ -0,0 +1,100 @@
...
+
+impl Tolls {
+    const MINT_ACCOUNT_INDEX_PAY_TOLL_INSTRUCTION: usize = 4;
+
+    pub async fn validate_toll_payment(
+        &self,
+        transaction: &VersionedTransaction,
+        chain_index: &ChainIndex,
+    ) -> Result<(), (StatusCode, String)> {
+        let tollbooth_instructions: Vec<(usize, &CompiledInstruction)> = transaction
```
anihamde: are we planning on releasing any SDK support for tollbooth? if so, could plug that in here to parse instead of manually checking

Tiltfile
```diff
@@ -42,6 +44,12 @@ local_resource(
     resource_deps=["svm-localnet"],
 )
 
+local_resource(
+    "setup-toll-recipient-ata",
```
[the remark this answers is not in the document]
anihamde: alternatively, we should consider how to enable ATA creation within the paymaster ahead of tollbooth payment. maybe have an idempotent ATA creation within the tollbooth program? that should be cheap in terms of CUs

services/paymaster/src/constraint/mod.rs
```diff
@@ -73,26 +76,18 @@ pub struct ContextualDomainKeys {
...
-        let program_id = instruction.program_id(static_accounts);
-        if program_id == program_to_match {
-            return Ok(true);
-        }
-    } else {
-        anyhow::bail!("Instruction index {instruction_index} out of bounds");
-    }
+impl VariationOrderedInstructionConstraints {
+    const IGNORED_PROGRAMS: &[Pubkey] =
+        &[solana_compute_budget_interface::id(), TOLLBOOTH_PROGRAM_ID];
```
cprussin: I'm going to take this offline but I feel uneasy about the idea of adding a set of instructions that are globally unconstrained

services/paymaster/src/constraint/mod.rs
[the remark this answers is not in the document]
cprussin: 100% agree on this point

services/paymaster/src/constraint/tolls.rs
```diff
@@ -0,0 +1,100 @@
...
+                ix.program_id(transaction.message.static_account_keys()) == &TOLLBOOTH_PROGRAM_ID
+            })
+            .collect();
+
+        match (self, tollbooth_instructions.as_slice()) {
+            (Tolls::Free, []) => Ok(()),
+            (Tolls::Free, _) => Err((
+                StatusCode::BAD_REQUEST,
+                "The paymaster for this domain does not require a toll payment".to_string(),
+            )),
```
cprussin: I don't think this should error; it does not hurt for the toll to be sent even when it isn't strictly required and it makes rollouts of changes easier, because it removes the requirement that the frontend change is synced to the paymaster config change.

Even if the paymaster were to e.g. expose an api that indicates current tolls, there's still a race condition where a tx created before the paymaster config was changed would fail if landed after the paymaster config changed.

IMO as long as the tx pays the requested toll _or more_, there shouldn't be a reason to stop the tx from landing.

services/paymaster/src/constraint/tolls.rs
```diff
@@ -0,0 +1,100 @@
...
+                        transaction,
+                        tollbooth_instruction,
+                        *instruction_index,
+                        Self::MINT_ACCOUNT_INDEX_PAY_TOLL_INSTRUCTION,
+                    )
+                    .await?;
+
+                if tolls
+                    .iter()
+                    .any(|toll| toll.mint == mint && toll.amount == amount)
```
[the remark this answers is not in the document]
cprussin: I think changing to `<=` is important for the reasons I mentioned in the last comment, as long as a tx is paying at least as much toll as requested or more, not requiring the toll to be exact makes rolling out changes easier since the UI does not have to be precisely synced to the paymaster config.

services/paymaster/src/constraint/tolls.rs
```diff
@@ -0,0 +1,100 @@
...
+                    ))
+            }
+            (Tolls::Fixed(_), []) => Err((
+                StatusCode::BAD_REQUEST,
+                "The paymaster for this domain requires a toll payment".to_string(),
+            )),
+            (Tolls::Fixed(_), [_, ..]) => Err((
+                StatusCode::BAD_REQUEST,
+                "This transaction contains multiple tollbooth instructions".to_string(),
+            )),
```
cprussin: As long as the sum of the tolls is at least the set toll amount, do we care if it's split in multiple ixs?

services/paymaster/src/api.rs
```diff
@@ -230,6 +232,10 @@ async fn sponsor_and_send_handler(
     let mut transaction: VersionedTransaction = bincode::deserialize(&transaction_bytes)
         .map_err(|_| (StatusCode::BAD_REQUEST, "Failed to deserialize transaction"))?;
 
+    domain_state
+        .tolls
+        .validate_toll_payment(&transaction, &state.chain_index)
+        .await?;
```
cprussin: Similar to the recent conversation about parse, don't validate -- is there any way we can represent that this validation has happened within the type system?

PR #238, opened 2025-10-17
chore: restrict token limits to tokens without a freeze authority

guibescos:


programs/session-manager/src/token/approve.rs
```diff
@@ -82,6 +82,11 @@ pub fn convert_remaning_accounts_and_token_limits_to_pending_approvals<'a, 'info
             );
 
             let mint_data = Mint::try_deserialize(&mut mint_account.data.borrow().as_ref())?;
+            require!(
+                mint_data.freeze_authority.is_none(),
```
anihamde: this is fine as a stopgap while most tokens have no freeze authority. but some mainstream tokens on solana do, so we will eventually need to come up with a better solution. i would add a TODO here

PR #240, opened 2025-10-20
fix: fix paymaster prop not being optional

cprussin:
This commit fixes an issue where in #234 , I did not properly mark the `paymaster` prop of the `<FogoSessionProvider>` as optional.  As a result consumers were required to define a `paymaster` prop unnecessarily.

### DIFF
--- a/packages/sessions-sdk-react/src/components/session-provider.tsx
+++ b/packages/sessions-sdk-react/src/components/session-provider.tsx
@@ -91,7 +91,7 @@ type Props = ConstrainedOmit<
   privacyPolicyUrl?: string | undefined;
 } & (
     | {
-        paymaster: ComponentProps<typeof SessionProvider>["paymaster"];
+        paymaster?: ComponentProps<typeof SessionProvider>["paymaster"];
         sendToPaymaster?: undefined;
         sponsor?: undefined;
       }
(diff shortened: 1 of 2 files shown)


packages/sessions-sdk-react/src/components/session-provider.tsx
```diff
@@ -91,7 +91,7 @@ type Props = ConstrainedOmit<
   privacyPolicyUrl?: string | undefined;
 } & (
     | {
-        paymaster: ComponentProps<typeof SessionProvider>["paymaster"];
+        paymaster?: ComponentProps<typeof SessionProvider>["paymaster"];
         sendToPaymaster?: undefined;
```
guibescos: why did you choose to include `sendToPaymaster`, `sponsor` when `paymaster` is provided and respectively `paymaster` when `sendToPaymaster` is provided

wouldn't 
```ts
{ paymaster?: ComponentProps<typeof SessionProvider>["paymaster"] ;} 
| {
        sendToPaymaster: NonNullable<
          ComponentProps<typeof SessionProvider>["sendToPaymaster"]
        >;
        sponsor: PublicKey | string;
      }
```
make more sense?

guibescos (approved):

packages/sessions-sdk-react/src/components/session-provider.tsx
guibescos: the argument about not needing awkward code to narrow the type makes sense to me

PR #245, opened 2025-10-21
chore: add ledger

guibescos:
Add support for ledger in the scripts

### DIFF
--- a/scripts/src/ledger.ts
+++ b/scripts/src/ledger.ts
@@ -12,8 +12,7 @@ export function parseDerivationPath(source: string): {
   derivationAccount?: number;
   derivationChange?: number;
 } {
-  const params = new URLSearchParams(new URL(source).searchParams);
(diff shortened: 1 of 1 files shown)


scripts/src/ledger.ts
```diff
@@ -0,0 +1,177 @@
...
+import { getDerivationPath } from "@solana/wallet-adapter-ledger";
+import type { Transaction } from "@solana/web3.js";
+import { VersionedTransaction, PublicKey } from "@solana/web3.js";
+
+export function parseDerivationPath(source: string): {
+  derivationAccount?: number;
+  derivationChange?: number;
+} {
+  const params = new URLSearchParams(new URL(source).searchParams);
+  const key = params.get("key");
```
cprussin: I think this is redundant; just do `const key = new URL(source).searchParams.get("key")`

PR #246, opened 2025-10-22
fix: paymaster pubsub robustness

anihamde:
We noticed some issues around paymaster uptime in production and believe this was mostly due to reliability issues around websocket connection via pubsub. This PR addresses that by adding reconnect logic to the pubsub client. It also adds some logging and span observability.

### DIFF
--- a/services/paymaster/src/rpc.rs
+++ b/services/paymaster/src/rpc.rs
@@ -294,7 +294,7 @@ async fn subscribe_and_wait_for_signature(
 
     result
         .map_err(|_| {
-            tracing::error!("Timeout while waiting for transaction confirmation for {signature}",);
+            tracing::error!("Timeout while waiting for transaction confirmation for {signature}");
             (
                 StatusCode::GATEWAY_TIMEOUT,
                 "Unable to confirm transaction".to_string(),


services/paymaster/src/rpc.rs
```diff
@@ -153,41 +156,111 @@ impl ChainIndex {
...
+            tracing::Span::current().record("result", "unconfirmed");
+            Err(e.into())
+        }
+    }
+}
+
+pub const SIGNATURE_SUBSCRIPTION_ERROR: &str = "Failed to subscribe to signature";
+
+fn is_subscription_error(status_code: StatusCode, error_string: &str) -> bool {
+    status_code == StatusCode::BAD_GATEWAY && error_string == SIGNATURE_SUBSCRIPTION_ERROR
```
anihamde: i don't love the string matching here, but i didn't want to change the error status code or rearchitect the error types too much just for this

services/paymaster/src/rpc.rs
```diff
@@ -153,41 +156,109 @@ impl ChainIndex {
...
+        }
+        Err(e) if is_subscription_error(e.0, &e.1) => {
+            tracing::warn!("WebSocket subscription failed, attempting reconnection and retry");
+            pubsub.reconnect_pubsub().await?;
+            let new_guard = pubsub.client.load();
+            let retry_result = confirm_transaction(&new_guard, signature, commitment)
+                .await
+                .map_err(|e| e.into());
+
+            match retry_result {
```
guibescos: this nested retry doesn't feel great. It leads to the duplication of:
- tracing::Span::current().record("result", confirmation.status_string().as_str());
- tracing::Span::current().record("result", "unconfirmed");

services/paymaster/src/rpc.rs
```diff
@@ -153,41 +156,110 @@ impl ChainIndex {
...
 #[tracing::instrument(
     skip_all,
-    fields(tx_hash = %signature)
+    name = "confirm_transaction",
+    fields(
+        tx_hash = %signature,
+        result
+    )
 )]
+async fn confirm_transaction_with_reconnect(
```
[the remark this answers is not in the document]
guibescos: I feel like conceptually, you may want all existing streams to terminate whenever you deem the pubsub stale or disconnected and replace it by a new one, I don't know if this makes anything easier.

services/paymaster/src/rpc.rs
```diff
@@ -181,28 +199,74 @@ pub const CONFIRMATION_TIMEOUT: Duration = Duration::from_secs(60);
...
+}
+
+#[async_recursion::async_recursion]
+async fn subscribe_and_wait_for_signature(
+    pubsub: &PubsubClientWithReconnect,
+    signature: Signature,
+    commitment: Option<CommitmentConfig>,
+    reconnect: bool,
+) -> Result<solana_client::rpc_response::ProcessedSignatureResult, (StatusCode, String)> {
+    let pubsub_client = pubsub.client.load();
```
guibescos: Out of curiosity, what is the difference between load and load_all?

guibescos (approved):

PR #252, opened 2025-10-24
feat: bridging intent transfer

anihamde:
This PR adds logic to the intent transfer program to enable bridging transfers via Wormhole NTT.

TODOs
- [ ] testing via unit tests/Tilt
- [ ] configure lookup tables for sizing

### DIFF
--- a/.github/workflows/cargo-test.yaml
+++ b/.github/workflows/cargo-test.yaml
@@ -21,7 +21,7 @@ jobs:
           echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
 
       - name: Build Solana programs
-        run: cargo +1.87.0 build-sbf
+        run: cargo build-sbf
 
       - name: Test
-        run: cargo +1.87.0 test
+        run: cargo test
--- a/programs/intent-transfer/src/error.rs
+++ b/programs/intent-transfer/src/error.rs
@@ -39,11 +39,9 @@ pub enum IntentTransferError {
     #[msg("The recipient address could not be parsed as a valid address")]
     InvalidRecipientAddress,
     #[msg("The provided to chain ID is unsupported")]
-    InvalidToChainId,
+    UnsupportedToChainId,
     #[msg("The provided Ntt manager for the given mint is invalid")]
     InvalidNttManager,
-    #[msg("Invalid ProgramData account")]
-    InvalidProgramData,
     #[msg("Unauthorized: only upgrade authority can call this")]
     Unauthorized,
 }
--- a/programs/intent-transfer/src/lib.rs
+++ b/programs/intent-transfer/src/lib.rs
@@ -56,11 +56,6 @@ pub mod intent_transfer {
     pub fn register_ntt_config<'info>(
         ctx: Context<'_, '_, '_, 'info, RegisterNttConfig<'info>>,
     ) -> Result<()> {
-        verify_upgrade_authority(
-            ctx.program_id,
-            &ctx.accounts.program_data,
-            &ctx.accounts.update_authority,
-        )?;
         ctx.accounts.expected_ntt_config.manager = ctx.accounts.ntt_manager.key();
         Ok(())
     }
@@ -68,7 +63,7 @@ pub mod intent_transfer {
 
 #[derive(Accounts)]
 pub struct RegisterNttConfig<'info> {
-    #[account(mut)]
+    #[account(mut, address = program_data.upgrade_authority_address.ok_or(IntentTransferError::Unauthorized)?)]
     pub update_authority: Signer<'info>,
 
     pub mint: Account<'info, Mint>,
@@ -85,7 +80,7 @@ pub struct RegisterNttConfig<'info> {
     /// CHECK: this is the address of the Ntt Manager program to register
     pub ntt_manager: UncheckedAccount<'info>,
 
-    /// this account's address is verified to make sure it corresponds to this program
+    #[account(address = bpf_loader_upgradeable::get_program_data_address(&crate::ID))]
     pub program_data: Account<'info, ProgramData>,
 
     pub system_program: Program<'info, System>,
@@ -345,7 +340,7 @@ impl<'info> BridgeNttTokens<'info> {
         )?;
 
         let to_chain_id_wormhole = convert_chain_id_to_wormhole(&to_chain_id)
-            .ok_or(IntentTransferError::InvalidToChainId)?;
+            .ok_or(IntentTransferError::UnsupportedToChainId)?;
 
         let transfer_args = cpi::ntt_manager::TransferArgs {
             amount,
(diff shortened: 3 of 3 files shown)


cprussin:
@anihamde something about this broke the Vercel builds, can you look into & fix it before merging?

Here's the vercel logs:

```
@fogo/sessions-idls:build:idl:example: Error: No such file or directory (os error 2)
@fogo/sessions-idls:build:idl:example:  ELIFECYCLE  Command failed with exit code 1.
@fogo/sessions-idls:build:idl:example: ERROR: command finished with error: command (/vercel/path0/packages/sessions-idls) /vercel/path0/.vercel/cache/corepack/shim/pnpm run build:idl:example exited (1)
@fogo/sessions-idls#build:idl:example: command (/vercel/path0/packages/sessions-idls) /vercel/path0/.vercel/cache/corepack/shim/pnpm run build:idl:example exited (1)
 ```

File: .github/workflows/release.yaml [hunk 1]
```diff
@@ -20,6 +20,12 @@ jobs:
           profile: minimal
           toolchain: nightly
 
+      - name: Install Solana
```
guibescos: do we need Solana here? afaik this workflow only uses anchor to generate the idls via cargo

.github/workflows/cargo-test.yaml
```diff
@@ -14,5 +14,22 @@ jobs:
         with:
           profile: minimal
           toolchain: 1.87.0
+
+      - name: Install solana cli
+        run: |
+          sh -c "$(curl -sSfL https://release.anza.xyz/v2.2.20/install)"
+          echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
+
+      - name: Install anchor cli
```
guibescos: nit: you could save some time here by not installing anchor and building with `cargo-build-sbf`

.github/workflows/cargo-test.yaml
```diff
@@ -14,5 +14,22 @@ jobs:
...
+        run: |
+          cargo install --git https://github.com/coral-xyz/anchor --tag v0.31.1 --locked avm
+          avm install 0.31.1
+          avm use 0.31.1
+
+      - name: Build Solana programs
+        run: |
+          export PATH="$HOME/.local/share/solana/install/active_release/bin:$PATH"
```
guibescos: seems like this was added to PATH already at line 21?

programs/intent-transfer/src/lib.rs
```diff
@@ -28,6 +36,368 @@ pub mod intent_transfer {
...
+#[derive(Accounts)]
+pub struct Ntt<'info> {
+    /// CHECK: Clock sysvar
+    pub clock: Sysvar<'info, Clock>,
+
+    /// CHECK: Rent sysvar
+    pub rent: Sysvar<'info, Rent>,
+
+    /// CHECK: checked in NTT manager program
+    pub ntt_manager: UncheckedAccount<'info>,
```
guibescos: do you need to check the address of this so it's not possible to CPI to an arbritary program?

Alternatively, `cpi::transfer_burn` can become `cpi::transfer_burn(ctx, args)` and hardcode program_id to the correct address of the program, and the CPI will fail if ntt_manager is not what it is.

programs/intent-transfer/src/lib.rs
```diff
@@ -28,6 +36,368 @@ pub mod intent_transfer {
...
+    #[account(mut, token::mint = mint)]
+    pub source: Account<'info, TokenAccount>,
+
+    #[account(
+        init_if_needed,
+        payer = sponsor,
+        seeds = [BRIDGE_NTT_INTERMEDIATE_SEED, source.key().as_ref()],
+        bump,
+        token::mint = mint,
+        token::authority = intent_transfer_setter
```
guibescos: nit: you could set the authority to `ntt_session_authority` right away here and you don't need to approve anymore

programs/intent-transfer/src/lib.rs
```diff
@@ -28,6 +36,368 @@ pub mod intent_transfer {
...
+    pub sponsor: Signer<'info>,
+
+    /// This signer exists solely to protect against MITM attacks whereby a malicious actor
+    /// intercepts a submitted transaction prior to any signatures being attached and changes
+    /// fields not directly encoded by the intent message. For example, such an attack could
+    /// alter the relay instructions in the bridging; attaching this signature prior to the
+    /// transaction leaving the client protects against that attack vector. Neither this signer
+    /// nor its signature are substantively required for this transaction. Note we do not need
+    /// to perform any checks on this signer; its mere presence is sufficient.
+    pub session_signer: Signer<'info>,
```
guibescos: this signer is unchecked, I don't think it prevents the MITM attack: I can still intercept the intent, replace this signer by a key I control and sign the new transaction with that key

programs/intent-transfer/tests/mocks/mock-ntt-manager/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+
+            let amount = u64::from_le_bytes([
+                instruction_data[8],
+                instruction_data[9],
+                instruction_data[10],
+                instruction_data[11],
+                instruction_data[12],
+                instruction_data[13],
+                instruction_data[14],
+                instruction_data[15],
```
guibescos: there has to be a better way to do this

programs/intent-transfer/src/lib.rs
anihamde: ah this is this way bc the NTT manager program is specific to the mint, and there are different NTT managers for different mints. The NTT manager program has a check to make sure that the right mint is passed in

programs/intent-transfer/src/lib.rs
```diff
@@ -576,6 +630,32 @@ fn verify_and_update_nonce(nonce: &mut Account<Nonce>, new_nonce: u64) -> Result
...
+    require_keys_eq!(
+        ntt_manager_key,
+        expected_ntt_config.manager,
+        IntentTransferError::InvalidNttManager
+    );
+    Ok(())
+}
+
+fn verify_upgrade_authority(
+    intent_transfer_program: &Pubkey,
```
guibescos: nit: you already have access to the program id via crate::ID

programs/intent-transfer/src/error.rs
```diff
@@ -40,6 +40,12 @@ pub enum IntentTransferError {
     InvalidRecipientAddress,
     #[msg("The provided to chain ID is unsupported")]
     InvalidToChainId,
+    #[msg("The provided Ntt manager for the given mint is invalid")]
+    InvalidNttManager,
+    #[msg("Invalid ProgramData account")]
+    InvalidProgramData,
```
guibescos: nit: it's more like a mismatch than an invalid. This happens when the program data provided is not the program data for the intent transfer program.

programs/intent-transfer/src/lib.rs
```diff
@@ -576,6 +630,32 @@ fn verify_and_update_nonce(nonce: &mut Account<Nonce>, new_nonce: u64) -> Result
...
+fn verify_ntt_manager(ntt_manager_key: Pubkey, expected_ntt_config: &Account<'_, ExpectedNttConfig>) -> Result<()> {
+    require_keys_eq!(
+        ntt_manager_key,
+        expected_ntt_config.manager,
+        IntentTransferError::InvalidNttManager
+    );
+    Ok(())
+}
+
+fn verify_upgrade_authority(
```
guibescos: nit: could write these as 
   `#[account(address = bpf_loader_upgradeable::get_program_data_address(&crate::ID))]`
and:
    `#[account(mut, address = program_data.upgrade_authority_address.ok_or(IntentTransferError::Unauthorized)?)]`

programs/intent-transfer/src/lib.rs
```diff
@@ -2,23 +2,31 @@
...
+pub mod bridge_message;
+pub mod cpi;
 pub mod error;
 mod message;
 
 const INTENT_TRANSFER_SEED: &[u8] = b"intent_transfer";
 const NONCE_SEED: &[u8] = b"nonce";
 
+const BRIDGE_NTT_INTERMEDIATE_SEED: &[u8] = b"bridge_ntt_intermediate";
+const BRIDGE_NTT_NONCE_SEED: &[u8] = b"bridge_ntt_nonce";
```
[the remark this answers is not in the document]
guibescos: I still feel we don't get a lot from having 2 nonce accounts

PR #257, opened 2025-10-30
feat: paymaster docker env & docs

alexcambose:
This PR adds a docker file that spins up a local pg db and a pg admin UI.
Also I've updated the readme now that we've changed the way the paymaster loads the config.
Screenshot 2025-10-30 at 13 29 45

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,8 +1,8 @@
 DATABASE_URL="postgres://paymaster:paymaster@localhost:5432/paymaster"
-OTEL_EXPORTER_OTLP_ENDPOINT=
-RPC_WS_HTTP=
-CONFIG_FILE=
-MNEMONIC_FILE = "./tilt/secrets/mnemonic"
-RPC_URL_HTTP = "https://testnet-alt.fogo.io"
-LISTEN_ADDRESS = "0.0.0.0:4000"
-DB_REFRESH_INTERVAL_SECONDS=10
+# MNEMONIC_FILE = "./tilt/secrets/mnemonic"
+# LISTEN_ADDRESS
+# DB_REFRESH_INTERVAL_SECONDS=
+# OTEL_EXPORTER_OTLP_ENDPOINT=
+# RPC_WS_HTTP=
+# CONFIG_FILE=
+# RPC_URL_HTTP=
--- a/Tiltfile
+++ b/Tiltfile
@@ -76,7 +76,7 @@ local_resource(
 
 local_resource(
     "paymaster",
-    serve_cmd="cargo run --bin fogo-paymaster migrate && cargo run --bin fogo-paymaster run",
+    serve_cmd="cargo run --bin fogo-paymaster migrate && cargo run --bin fogo-paymaster run --db-url postgres://paymaster:paymaster@localhost:5432/paymaster",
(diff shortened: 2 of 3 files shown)


.env.example
```diff
@@ -1,8 +1,8 @@
-DATABASE_URL=
+DATABASE_URL="postgres://paymaster:paymaster@localhost:5432/paymaster"
```
alexcambose: might come in handy if we add these in the example env

.env.example
```diff
@@ -1,8 +1,8 @@
...
 OTEL_EXPORTER_OTLP_ENDPOINT=
-MNEMONIC_FILE=
-LISTEN_ADDRESS=
-RPC_URL_HTTP=
 RPC_WS_HTTP=
 CONFIG_FILE=
```
guibescos: I think the empty string here are not good here because the paymaster reads them as empty strings (and not `None`) and crashes

.env.example
```diff
@@ -1,8 +1,8 @@
...
-DB_REFRESH_INTERVAL_SECONDS=
+MNEMONIC_FILE = "./tilt/secrets/mnemonic"
+RPC_URL_HTTP = "https://testnet-alt.fogo.io"
+LISTEN_ADDRESS = "0.0.0.0:4000"
+DB_REFRESH_INTERVAL_SECONDS=10
```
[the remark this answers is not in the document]
alexcambose: that's a good point

.env.example
alexcambose: hmm yeah I wanted to list out the available env vars but without necessarily assigning something specific.
I'll comment out the empty ones

.env.example
```diff
@@ -1,8 +1,8 @@
...
-CONFIG_FILE=
-DB_REFRESH_INTERVAL_SECONDS=
+DATABASE_URL="postgres://paymaster:paymaster@localhost:5432/paymaster"
+# MNEMONIC_FILE = "./tilt/secrets/mnemonic"
+# LISTEN_ADDRESS
+# DB_REFRESH_INTERVAL_SECONDS=
+# OTEL_EXPORTER_OTLP_ENDPOINT=
+# RPC_WS_HTTP=
+# CONFIG_FILE=
+# RPC_URL_HTTP=
```
guibescos: now that Tilt doesn't need this file, it doesn't feel very useful
maybe in the future it'd be more useful to have something like
.env.testnet.example
.env.mainnet.example

guibescos (approved):

.env.example
```diff
@@ -1,8 +1,8 @@
...
-RPC_WS_HTTP=
-CONFIG_FILE=
-DB_REFRESH_INTERVAL_SECONDS=
+DATABASE_URL="postgres://paymaster:paymaster@localhost:5432/paymaster"
+# MNEMONIC_FILE = "./tilt/secrets/mnemonic"
+# LISTEN_ADDRESS
+# DB_REFRESH_INTERVAL_SECONDS=
+# OTEL_EXPORTER_OTLP_ENDPOINT=
+# RPC_WS_HTTP=
+# CONFIG_FILE=
```
guibescos: also does this file need to be at the root level

PR #258, opened 2025-10-30
test: paymaster load testing

anihamde:
This PR adds a load testing script to evaluate the performance of the paymaster under simulated load. The script can be configured to evaluate against local or production paymasters. It tracks end-to-end latency and confirmation rate.

### DIFF
--- a/scripts/paymaster-load-test/src/metrics.rs
+++ b/scripts/paymaster-load-test/src/metrics.rs
@@ -42,33 +42,25 @@ impl LoadTestMetrics {
     pub fn record_success(&self, validity_type: ValidityType, latency: Duration) {
         let latency_us = latency.as_micros() as u64;
 
-        {
-            let mut global = self.0.entry(None).or_default();
-            global.succeeded += 1;
-            let _ = global.e2e_latency.record(latency_us);
-        }
+        let mut global = self.0.entry(None).or_default();
+        global.succeeded += 1;
+        let _ = global.e2e_latency.record(latency_us);
 
-        {
-            let mut entry = self.0.entry(Some(validity_type)).or_default();
-            entry.succeeded += 1;
-            let _ = entry.e2e_latency.record(latency_us);
-        }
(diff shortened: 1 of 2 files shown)


scripts/paymaster-load-test/README.md
```diff
@@ -0,0 +1,67 @@
+# Paymaster Load Test
```
guibescos: let's move the typescript package that lives in `scripts` to `scripts/ts`

scripts/paymaster-load-test/src/main.rs
```diff
@@ -0,0 +1,136 @@
...
+        let current_rate = requests_delta as f64 / interval.as_secs_f64();
+        let success_rate = metrics.success_rate();
+
+        tracing::info!(
+            "[{:3}s] Requests: {} (+{} @ {:.1} req/s) | Success: {:.1}%",
+            (i + 1) * 5,
+            current_requests,
+            requests_delta,
+            current_rate,
+            success_rate * 100.0
```
guibescos: this success rate is a little bit confusing since at this point we haven't received a response for some requests, but that doesn't mean they have failed

guibescos (commented):
This code is very nice, main comment is moving the Typescript scripts for repo hygiene, let me know if you need help with that

scripts/paymaster-load-test/src/metrics.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    pub fn record_success(&self, validity_type: ValidityType, latency: Duration) {
+        let latency_us = latency.as_micros() as u64;
+
+        {
+            let mut global = self.0.entry(None).or_default();
+            global.succeeded += 1;
+            let _ = global.e2e_latency.record(latency_us);
+        }
```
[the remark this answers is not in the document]
anihamde: these were originally in place to prevent some conflicts and deadlocking. i don't think it's a problem anymore, we can get rid of them

PR #262, opened 2025-10-31
refactor: Improve paymaster validation

guibescos:
The goal of this PR is to improve the type system in the paymaster transaction validation based on feedback from https://github.com/fogo-foundation/fogo-sessions/pull/235.

- we introduce a type `TransactionToValidate` that we use to "parse" the transaction right away, separating the compute instructions from the substantive instructions and in the future the tollbooth instructions
- we only compute the gas spend once instead of once for validation and one for metrics
- we clean up out of bounds checks when accessing in the validation of instructions against contraints by using a Deque
- we introduce InstructionWithIndex to avoid needing to pass the instruction AND the index everywhere

### DIFF
--- a/services/paymaster/src/constraint/mod.rs
+++ b/services/paymaster/src/constraint/mod.rs
@@ -117,7 +117,8 @@ impl VariationOrderedInstructionConstraints {
         contextual_domain_keys: &ContextualDomainKeys,
         chain_index: &ChainIndex,
     ) -> Result<(), (StatusCode, String)> {
-        let mut substantive_instructions_iterator = transaction.substantive_instructions.iter().peekable();
+        let mut substantive_instructions_iterator =
+            transaction.substantive_instructions.iter().peekable();
 
         // Note: this validation algorithm is technically incorrect, because of optional constraints.
         // E.g. instruction i might match against both constraint j and constraint j+1; if constraint j


services/paymaster/src/constraint/gas.rs
```diff
@@ -0,0 +1,100 @@
...
+use borsh::BorshDeserialize;
+use solana_compute_budget_interface::ComputeBudgetInstruction;
+use solana_message::compiled_instruction::CompiledInstruction;
+use solana_message::VersionedMessage;
+use solana_pubkey::Pubkey;
+use solana_sdk_ids::{ed25519_program, secp256k1_program, secp256r1_program};
+use solana_transaction::versioned::VersionedTransaction;
+
+pub const LAMPORTS_PER_SIGNATURE: u64 = 5000;
+pub const DEFAULT_COMPUTE_UNIT_LIMIT: u64 = 200_000;
```
cprussin: I don't think these need to be `pub`

services/paymaster/src/constraint/gas.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    let priority_fee = cu_limit
+        .unwrap_or(DEFAULT_COMPUTE_UNIT_LIMIT)
+        .saturating_mul(micro_lamports_per_cu.unwrap_or(0))
+        / 1_000_000;
+    Ok(priority_fee)
+}
+
+/// The Solana precompile programs that verify signatures.
+pub const PRECOMPILE_SIGNATURE_PROGRAMS: &[Pubkey] = &[
```
cprussin: Also does not need `pub`

PR #270, opened 2025-11-09
feat: add script to register ntt manager

cprussin:
This commit also includes some cleanup for all of the scripts

### DIFF
--- a/scripts/src/add-program-id-to-domain-registry.ts
+++ b/scripts/src/add-program-id-to-domain-registry.ts
@@ -8,7 +8,10 @@ import { anchorOptions, createAnchorProvider } from "./anchor-options.js";
 
 export const main = async (argv: string[] = hideBin(process.argv)) => {
   const args = await yargs(argv)
-    .command("* <domain> <program-id>", "command")
+    .command(
+      "* <domain> <program-id>",
+      "Add the given program ID to the whitelist of programs for the given domain",
+    )
     .options(anchorOptions)
     .positional("domain", {
       type: "string",
@@ -22,7 +25,7 @@ export const main = async (argv: string[] = hideBin(process.argv)) => {
     })
     .parse();
 
-  const program = new DomainRegistryProgram(await createAnchorProvider(args));
+  const program = new DomainRegistryProgram(createAnchorProvider(args));
 
   const { config: configPubkey } = await program.methods.initialize().pubkeys();
 
--- a/scripts/src/anchor-options.ts
+++ b/scripts/src/anchor-options.ts
@@ -17,14 +17,19 @@ export const anchorOptions = {
     type: "string",
     description: "Filepath to a keypair",
     demandOption: true,
-    coerce: (keypair: string) => parseSignerSource(keypair),
+    coerce: (keypair: string): Awaited<ReturnType<typeof parseSignerSource>> =>
+      // @ts-expect-error yargs types are incorrect.  yargs will await the
+      // coerce but the typescript types do not indicate so, so we explicitly
+      // override yargs here to indicate that the return value is indeed
+      // awaited.
+      parseSignerSource(keypair),
   },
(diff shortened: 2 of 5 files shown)


programs/intent-transfer/src/lib.rs
```diff
@@ -86,6 +86,7 @@ pub struct RegisterNttConfig<'info> {
     pub ntt_manager: UncheckedAccount<'info>,
 
     /// this account's address is verified to make sure it corresponds to this program
+    #[account(seeds = [crate::ID.as_ref()], bump, seeds::program = bpf_loader_upgradeable::ID)]
```
cprussin: @anihamde I think this should be included in your commit as per @guibescos 's feedback but I added it here so I could write the scripts without manually specifying this account

scripts/src/ledger.ts
```diff
@@ -1,12 +1,24 @@
...
   TransportStatusError,
 } from "@ledgerhq/hw-transport";
 import TransportNodeHid from "@ledgerhq/hw-transport-node-hid";
 import { getDerivationPath } from "@solana/wallet-adapter-ledger";
 import type { Transaction } from "@solana/web3.js";
-import { VersionedTransaction, PublicKey } from "@solana/web3.js";
+import { Keypair, VersionedTransaction, PublicKey } from "@solana/web3.js";
+
+export const parseSignerSource = async (source: string) =>
+  source.startsWith("usb://ledger")
```
benduran: q: should we do a case-insensitive check via a RegExp or `String.prototype..toLowerCase()` first, or is it safe to rely on the string already being lowercased?

benduran (approved):
left some comments and one suggestion. nothing major. take a look whenever you have time, otherwise, LGTM.

it's okay

scripts/src/anchor-options.ts
```diff
@@ -0,0 +1,29 @@
...
+  },
+} as const;
+
+export const createAnchorProvider = async ({
+  url,
+  keypair,
+}: {
+  url: string;
+  keypair: ReturnType<typeof parseSignerSource>;
+}) => new AnchorProvider(new Connection(url), await keypair);
```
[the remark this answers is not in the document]
cprussin: Yeah unfortunately this is wonky but not really something we can easily fix because we're using `coerce` to call `parseSignerSource` and yargs does not await `coerce` responses.  I figured it's better to coerce in yargs so the arg at the type level can't be misused, with the tradeoff that the api is a bit wonky and we're passing around a promise, than it is to just leave the keypair args as an arbitrary string and then require any users to manually convert (and potentially not do it correctly since it's not a trivial `new Keypair(str)` in this case)

scripts/src/register-ntt-manager.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const provider = await createAnchorProvider(args);
+  const program = new IntentTransferProgram(provider);
+
+  await program.methods
+    .registerNttConfig()
+    .accounts({
+      mint: args.mint,
+      nttManager: args.nttManager,
+      updateAuthority: provider.wallet.publicKey,
```
guibescos: you don't need to provide `updateAuthority`, it will default to the signer of this provider

guibescos (approved):

scripts/src/add-program-id-to-domain-registry.ts
```diff
@@ -1,71 +1,28 @@
...
-      alias: "u",
-      type: "string",
-      description: "URL for Solana's JSON RPC or moniker",
-    })
-    .option("keypair", {
-      alias: "k",
-      type: "string",
-      description: "Filepath to a keypair",
-    })
+    .command("* <domain> <program-id>", "command")
```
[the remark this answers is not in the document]
cprussin: lol good call, meant to go back and update this and forgot.  Will do it now

PR #272, opened 2025-11-10
feat: add script to register ntt manager

cprussin:
I accidentally merged https://github.com/fogo-foundation/fogo-sessions/pull/270 to the wrong branch.  This commit is the same, but now pointing to main instead.


programs/intent-transfer/src/lib.rs
```diff
@@ -80,7 +80,7 @@ pub struct RegisterNttConfig<'info> {
     /// CHECK: this is the address of the Ntt Manager program to register
     pub ntt_manager: UncheckedAccount<'info>,
 
-    #[account(address = bpf_loader_upgradeable::get_program_data_address(&crate::ID))]
+    #[account(seeds = [crate::ID.as_ref()], bump, seeds::program = bpf_loader_upgradeable::ID)]
```
anihamde: can we undo this?

PR #275, opened 2025-11-10
chore: breakdown big file

guibescos:
Break-down the big `lib.rs` in intent-transfer


programs/intent-transfer/src/lib.rs
```diff
@@ -2,34 +2,19 @@
...
+mod verify;
 
-const INTENT_TRANSFER_SEED: &[u8] = b"intent_transfer";
-const NONCE_SEED: &[u8] = b"nonce";
-
-const BRIDGE_NTT_INTERMEDIATE_SEED: &[u8] = b"bridge_ntt_intermediate";
-const BRIDGE_NTT_NONCE_SEED: &[u8] = b"bridge_ntt_nonce";
+use bridge::config::processor::register_ntt_config::*;
+use bridge::processor::bridge_ntt_tokens::*;
+use internal::processor::send_tokens::*;
```
guibescos: anchor is unhappy if I don't do this

programs/intent-transfer/src/bridge/config/ntt_config.rs
```diff
@@ -0,0 +1,22 @@
+use crate::error::IntentTransferError;
```
anihamde: i feel like all the config stuff can live in config.rs? tree feels unnecessarily deep otherwise

programs/intent-transfer/src/bridge/processor/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod bridge_ntt_tokens;
```
anihamde: why do we have a one file subfolder? why not just put `bridge_ntt_tokens.rs` into `processor.rs`?

programs/intent-transfer/src/internal/processor/send_tokens.rs
```diff
@@ -0,0 +1,119 @@
+use crate::{
```
anihamde: instead of `internal`, maybe `onchain`/`cischain`/`intrachain`?

programs/intent-transfer/src/bridge/config/ntt_config.rs
guibescos: That would break the pattern of having a "processor" file with just the `context` and the `impl` and nothing else for the `register_ntt_config` instruction

PR #280, opened 2025-11-11
feat: send tokens with fee

guibescos:
adds fees for sending and briding out

### DIFF
--- a/packages/sessions-sdk-react/src/components/send-token-page.tsx
+++ b/packages/sessions-sdk-react/src/components/send-token-page.tsx
@@ -205,8 +205,15 @@ export const SendTokenPage = ({
               <MaxButton
                 feeConfig={feeConfig}
                 excludeFromTabOrder={showScanner}
-                setAmount={(fee) => {
-                  setAmount(amountToString(amountAvailable - fee, decimals));
+                setAmount={({ fee, mint: feeMint }) => {
+                  setAmount(
+                    amountToString(
+                      feeMint.equals(tokenMint)
+                        ? amountAvailable - fee
+                        : amountAvailable,
+                      decimals,
+                    ),
+                  );
                 }}
               />
             </Suspense>
@@ -264,15 +271,15 @@ const MaxButton = ({
 }: {
   feeConfig: ReturnType<typeof getTransferFee>;
   excludeFromTabOrder: boolean;
-  setAmount: (fee: bigint) => void;
+  setAmount: (feeConfig: Awaited<ReturnType<typeof getTransferFee>>) => void;
 }) => {
-  const { fee } = use(feeConfig);
+  const resolvedConfig = use(feeConfig);
   return (
     <Link
       excludeFromTabOrder={excludeFromTabOrder}
       className={styles.action ?? ""}
       onPress={() => {
-        setAmount(fee);
+        setAmount(resolvedConfig);
       }}
     >
       Max
--- a/packages/sessions-sdk-react/src/components/withdraw-page.tsx
+++ b/packages/sessions-sdk-react/src/components/withdraw-page.tsx
@@ -172,10 +172,12 @@ const WithdrawForm = ({
               {...(props.isLoading
                 ? { isPending: true }
                 : {
-                    setAmount: (fee) => {
+                    setAmount: ({ fee, mint: feeMint }) => {
                       setAmount(
                         amountToString(
-                          props.amountAvailable - fee,
+                          feeMint.equals(USDC.chains[network].fogo.mint)
+                            ? props.amountAvailable - fee
+                            : props.amountAvailable,
                           USDC.decimals,
                         ),
                       );
@@ -221,9 +223,14 @@ const MaxButton = ({
   feeConfig: ReturnType<typeof getBridgeOutFee>;
 } & (
   | { isPending: true }
-  | { isPending?: false | undefined; setAmount: (fee: bigint) => void }
+  | {
+      isPending?: false | undefined;
+      setAmount: (
+        feeConfig: Awaited<ReturnType<typeof getBridgeOutFee>>,
+      ) => void;
+    }
 )) => {
-  const { fee } = use(feeConfig);
+  const resolvedConfig = use(feeConfig);
   return (
     <Link
       className={styles.action ?? ""}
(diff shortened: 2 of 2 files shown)


programs/intent-transfer/src/intrachain/processor/send_tokens_with_fee.rs
```diff
@@ -0,0 +1,88 @@
...
+};
+
+#[derive(Accounts)]
+pub struct SendTokensWithFee<'info> {
+    pub send_tokens: SendTokens<'info>,
+
+    #[account(mut, token::mint = fee_mint, token::authority = send_tokens.source.owner )]
+    pub fee_source: Account<'info, TokenAccount>,
+
+    #[account(init_if_needed, payer = send_tokens.sponsor, associated_token::mint = fee_mint, associated_token::authority = send_tokens.sponsor)]
```
cprussin: I don't think this should be `init_if_needed`; fees should never be paid to an uninitialized account so this seems like an error if it ever happens

scripts/src/register-ntt-manager.ts
```diff
@@ -28,6 +28,10 @@ export const main = async (argv: string[] = hideBin(process.argv)) => {
 
   await new IntentTransferProgram(createAnchorProvider(args)).methods
     .registerNttConfig()
-    .accounts({ mint: args.mint, nttManager: args.nttManager })
+    .accounts({
+      mint: args.mint,
+      nttManager: args.nttManager,
+      upgradeAuthority: { signer: undefined },
```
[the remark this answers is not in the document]
guibescos: anchor types have some weird behavior with the nested account context and I had to do this

programs/intent-transfer/src/config/state/fee_config.rs
```diff
@@ -0,0 +1,95 @@
...
+        self.verify_and_collect_fee(
+            self.bridging_out_fee,
+            args,
+            fee_amount,
+            fee_symbol_or_mint,
+            signer_seeds,
+        )
+    }
+
+    fn verify_and_collect_fee(
```
cprussin: The abstraction here seems incorrect to me, this shouldn't be a method in the `FeeConfig` impl.  No reference to the fee config (`self`) is used anywhere.

I think a better abstraction would be to make a trait with the method `fee_amount` and implement that trait on the two account structs, and then you can implement a default `verify_and_collect_fee` in terms of that `fee_amount`

cprussin (approved):
Minor note on the abstraction but otherwise LGTM

programs/intent-transfer/src/fees.rs
```diff
@@ -0,0 +1,59 @@
...
+            fee_metadata,
+            intent_transfer_setter,
+            token_program,
+        } = self.verify_and_collect_accounts();
+
+        verify_symbol_or_mint(&fee_symbol_or_mint, fee_metadata, fee_mint)?;
+        let intent_fee_amount = try_ui_amount_into_amount(intent_fee_amount, fee_mint.decimals)?;
+        require_gte!(
+            intent_fee_amount,
+            self.fee_amount(),
```
anihamde: nit: just call fee_amount() once and make a variable

programs/intent-transfer/src/config/processor/register_fee_config.rs
```diff
@@ -0,0 +1,28 @@
+use crate::config::access_control::*;
+use crate::config::state::fee_config::{FeeConfig, FEE_CONFIG_SEED};
+use anchor_lang::prelude::*;
+use anchor_spl::token::Mint;
+
+#[derive(Accounts)]
+pub struct RegisterFeeConfig<'info> {
```
anihamde: slight nit: perhaps this should be `UpdateFeeConfig`? to suggest that this can be called multiple times to update the fees

programs/intent-transfer/tests/bridge_ntt_tokens.rs
```diff
@@ -2,41 +2,52 @@ use anchor_lang::{
...
+    },
 };
 
 mod helpers;
 
+#[allow(clippy::too_many_arguments)]
 fn create_ntt_bridge_message(
     from_chain_id: &str,
+    to_chain_id: &str,
     token_symbol: &str,
```
anihamde: arguably can combine the symbol and amount into a single struct to pass in

anihamde (approved):

programs/intent-transfer/src/config/state/fee_config.rs
```diff
@@ -0,0 +1,10 @@
+use anchor_lang::prelude::*;
+
+pub const FEE_CONFIG_SEED: &[u8] = b"fee_config";
+
+#[account]
+#[derive(InitSpace)]
+pub struct FeeConfig {
+    pub ata_creation_fee: u64,
```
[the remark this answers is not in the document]
guibescos: I agree

packages/sessions-sdk-ts/src/index.ts
```diff
@@ -761,6 +761,55 @@ export type Session = {
   sessionInfo: NonNullable<z.infer<typeof sessionInfoSchema>>;
 };
 
+const USDC_MINT = {
+  [Network.Mainnet]: "uSd2czE61Evaf76RNbq4KPpXnkiL3irdzgLFUMe3NoG",
+  [Network.Testnet]: "ELNbJ1RtERV2fjtuZjbTscDekWhVzkQ1LjmiPsxp5uND",
+};
+const USDC_DECIMALS = 6;
```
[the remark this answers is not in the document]
cprussin: Yep, you're right, I'll refactor this a bit.  Why don't we shoot to do it in a follow up though because what I want to do will involve a bit more reorganization and I want to decouple it from this feature.

packages/sessions-sdk-react/src/components/send-token-page.tsx
```diff
@@ -180,15 +190,21 @@ export const SendTokenPage = ({
...
+                </Link>
+              }
             >
-              Max
-            </Link>
+              <MaxButton
+                feeConfig={feeConfig}
+                excludeFromTabOrder={showScanner}
+                setAmount={(fee) => {
+                  setAmount(amountToString(amountAvailable - fee, decimals));
```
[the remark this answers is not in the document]
cprussin: ah yes you're absolutely right

PR #282, opened 2025-11-11
fix: add wormholeProgram account

cprussin:


packages/sessions-sdk-ts/src/index.ts
```diff
@@ -981,6 +981,7 @@ const getNttPdas = async <N extends WormholeNetwork>(
     nttTokenAuthority: pdas.tokenAuthority(),
     payeeNttWithExecutor: quotePayeeAddress,
     transceiver: options.fromToken.transceiver,
+    wormholeProgram: coreBridgeContract,
```
anihamde: where does `coreBridgeContract` come from?

anihamde (approved):

packages/sessions-sdk-ts/src/index.ts
cprussin: There's a registry in wormhole that exports it:

```
import { contracts } from "@wormhole-foundation/sdk-base";

const coreBridgeContract = contracts.coreBridge.get(wh.network, "Fogo");
```

Where `wh.network` is `"Mainnet" | "Testnet"`

PR #288, opened 2025-11-12
feat(paymaster): add latency metrics

guibescos:


services/paymaster/src/metrics.rs
```diff
@@ -1,4 +1,5 @@
 use axum_prometheus::metrics;
+use std::time::Duration;
```
anihamde: nit: we could have a separate file for spend and latency metrics?

services/paymaster/src/metrics.rs
```diff
@@ -22,6 +23,48 @@ pub fn obs_send(domain: String, variation: String, result_confirmation: String)
     metrics::counter!(TRANSACTION_SEND_COUNT, &labels).increment(1);
 }
 
+pub const TRANSACTION_CONFIRMATION_NOTIFICATION_LATENCY: &str =
+    "paymaster_transaction_confirmation_notification_latency_seconds";
+pub const TRANSACTION_ACTUAL_CONFIRMATION_LATENCY: &str =
+    "paymaster_transaction_actual_confirmation_latency_seconds";
+
+pub const TRANSACTION_CONFIRMATION_LATENCY_BUCKETS: &[f64] =
```
anihamde: did you check if these buckets are appropriately sized? like if all of our latencies end up less than 0.5, then this isn't good

PR #292, opened 2025-11-13
feat: add explorer links to confirmation toasts

cprussin:


packages/sessions-sdk-react/src/components/explorer-link.tsx
```diff
@@ -0,0 +1,50 @@
...
+  txHash: string;
+};
+
+export const ExplorerLink = ({
+  children,
+  network,
+  txHash,
+  ...props
+}: Props) => {
+  const link = useMemo(() => mkLink(network, txHash), [network]);
```
benduran: nit: `mkLink()` looks to be a function with a simple switch statement, so we can likely drop the `useMemo()` to save on memory

packages/sessions-sdk-react/src/components/send-token-page.tsx
```diff
@@ -77,7 +78,10 @@ export const SendTokenPage = ({
         )
         .then((result) => {
           if (result.type === TransactionResultType.Success) {
-            toast.success("Tokens sent successfully!");
+            toast.success(
+              "Tokens sent successfully!",
```
anihamde: if the relevant paymaster doesn't have preflight might this tx fail on chain? in which case tokens were not actually sent

packages/sessions-sdk-react/src/components/withdraw-page.tsx
```diff
@@ -92,6 +93,7 @@ const WithdrawForm = ({
           if (result.type === TransactionResultType.Success) {
             toast.success(
               "Successful transferred tokens to your Solana wallet!",
+              <ExplorerLink network={network} txHash={result.signature} />,
```
anihamde: same thing here, what if this doesn't actually succeed on chain? e.g. wh issues/relayer downtime/etc

cprussin: See above, also guarded by `result.type === TransactionResultType.Success`

PR #295, opened 2025-11-14
[Feature] Add FTL support to paymaster

Dewyer:
# About
This PR adds support for using the custom FTL service / extension to the solana RPC's sendTransaction method.
it allows callers to send AND confirm in one http call to the service.

Support for this has been added under the optional ftl-url cli option, which if specific skips using the regular rpc url for submitting and confirming txs and isntead uses the specified ftl address.

### DIFF
--- a/services/paymaster/src/api.rs
+++ b/services/paymaster/src/api.rs
@@ -45,8 +45,6 @@ use tower_http::cors::{AllowHeaders, AllowMethods, AllowOrigin, CorsLayer};
 use tracing::Instrument;
 use utoipa_axum::{router::OpenApiRouter, routes};
 
-const RPC_POOL_SIZE: usize = 6;
-
 pub struct DomainState {
     pub domain_registry_key: Pubkey,
     pub sponsor: Keypair,
@@ -471,6 +469,10 @@ pub fn get_domain_state_map(domains: Vec<Domain>, mnemonic: &str) -> HashMap<Str
         )
         .collect::<HashMap<_, _>>()
 }
+
+/// How many RPC clients to create for both FTL and the regular RPC client for read operations
+const RPC_POOL_SIZE: usize = 6;
+
 pub async fn run_server(
     rpc_url_http: String,
     rpc_url_ws: String,
@@ -482,12 +484,7 @@ pub async fn run_server(
 
     let rpc = RpcClient::new_sender(
         rpc_http_sender,
-        RpcClientConfig {
-            commitment_config: CommitmentConfig {
-                commitment: CommitmentLevel::Processed,
-            },
-            confirm_transaction_initial_timeout: None,
-        },
+        RpcClientConfig::with_commitment(CommitmentConfig::processed()),
     );
 
     let rpc_sub_client = PubsubClient::new(&rpc_url_ws)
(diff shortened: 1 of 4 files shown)


services/paymaster/src/api.rs
```diff
@@ -460,20 +474,42 @@ pub fn get_domain_state_map(domains: Vec<Domain>, mnemonic: &str) -> HashMap<Str
...
-        CommitmentConfig {
-            commitment: CommitmentLevel::Processed,
+    let rpc_http_sender = PooledHttpSender::new(rpc_url_http, RPC_POOL_SIZE);
+
+    let rpc = RpcClient::new_sender(
+        rpc_http_sender,
+        RpcClientConfig {
+            commitment_config: CommitmentConfig {
+                commitment: CommitmentLevel::Processed,
+            },
```
guibescos: nit: I think this reads better
```suggestion
        RpcClientConfig::with_commitment(CommitmentLevel::Processed)
```

services/paymaster/src/rpc.rs
```diff
@@ -195,6 +196,110 @@ pub async fn send_and_confirm_transaction(
...
+        tracing::error!("Failed to serialize config: {e}");
+        ErrorResponse::from((
+            StatusCode::INTERNAL_SERVER_ERROR,
+            "Failed to serialize config",
+        ))
+    })?;
+
+    // Add the "confirm" parameter to the config
+    if let Some(obj) = config_value.as_object_mut() {
+        obj.insert("confirm".to_string(), json!(true));
```
guibescos: so to make sure I understand, the only reason we can't use `ftl_rpc.send_transaction` is because we need to add this to `RpcSendTransactionConfig`?

services/paymaster/src/pooled_http_sender.rs
```diff
@@ -0,0 +1,226 @@
...
+    async fn send(
+        &self,
+        request: RpcRequest,
+        params: serde_json::Value,
+    ) -> Result<serde_json::Value> {
+        let request_id = self.request_id.fetch_add(1, Ordering::Relaxed);
+        let method = format!("{request}");
+        let request_json = request.build_request_json(request_id, params).to_string();
+
+        let client = self.clients[(request_id as usize) % self.clients.len()].clone();
```
guibescos: nit: you could remove this cast by making request_id AtomicUsize

guibescos (commented):
Some minor comments

services/paymaster/src/pooled_http_sender.rs
```diff
@@ -0,0 +1,226 @@
...
+use solana_rpc_client_api::error_object::RpcErrorObject;
+use solana_transaction_error::TransactionError;
+use std::error::Error;
+use std::fmt::Write;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::sleep;
+
+pub const TRANSACTION_CONFIRMED_FAILED: i64 = -32604;
```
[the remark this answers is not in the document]
Dewyer: nope, this is part of the "extension" that does the confirming, more about it here:
https://github.com/dourolabs/fogo-transaction-lander/blob/main/README.md#error-handling

PR #296, opened 2025-11-14
feat: improve wh parsing

anihamde:
* alter program to construct ntt cpi args internally

* clippy + fmt

* le -> be

* correct calculation

* add test

* clippy+fmt

* comments on overflow


packages/sessions-sdk-ts/src/index.ts
```diff
@@ -935,6 +939,16 @@ export const bridgeOut = async (options: SendBridgeOutOptions) => {
   );
 };
 
+const getDestinationAtaExists = async (
+  context: SessionContext,
+  token: PublicKey,
+  wallet: PublicKey,
+) => {
```
anihamde: is this gonna return a bool? or like an optional object?

PR #298, opened 2025-11-14
fix: parse wh ntt signed quote bytes as big endian

anihamde:
This PR fixes the WH signed quote bytes to be parsed with numerics interpreted in big endian.

### DIFF
--- a/programs/intent-transfer/tests/bridge_ntt_tokens.rs
+++ b/programs/intent-transfer/tests/bridge_ntt_tokens.rs
@@ -10,14 +10,12 @@ use solana_signer::Signer;
 use solana_transaction::Transaction;
 use spl_token::solana_program::keccak;
 
-use intent_transfer::{
-    bridge::{
-        be::{U16BE, U64BE},
-        config::ntt_config::ExpectedNttConfig,
-        cpi::ntt_with_executor::{EXECUTOR_PROGRAM_ID, NTT_WITH_EXECUTOR_PROGRAM_ID},
-        message::convert_chain_id_to_wormhole,
-        processor::bridge_ntt_tokens::{BridgeNttTokensArgs, SignedQuote, SignedQuoteHeader},
-    },
+use intent_transfer::bridge::{
+    be::{U16BE, U64BE},
+    config::ntt_config::ExpectedNttConfig,
+    cpi::ntt_with_executor::{EXECUTOR_PROGRAM_ID, NTT_WITH_EXECUTOR_PROGRAM_ID},
+    message::convert_chain_id_to_wormhole,
+    processor::bridge_ntt_tokens::{BridgeNttTokensArgs, SignedQuote, SignedQuoteHeader},
 };
 
 mod helpers;


programs/intent-transfer/src/bridge/processor/bridge_ntt_tokens.rs
```diff
@@ -473,6 +470,124 @@ pub struct SignedQuote {
...
+        let mut quoter_address = [0u8; 20];
+        quoter_address.copy_from_slice(read_bytes!(20));
+
+        let mut payee_address = [0u8; 32];
+        payee_address.copy_from_slice(read_bytes!(32));
+
+        let source_chain = u16::from_be_bytes(
+            bytes[offset..offset + 2]
+                .try_into()
+                .map_err(|_| IntentTransferError::InvalidNttSignedQuote)?,
```
cprussin: why not `read_bytes!(2).try_into().map_err()` ?

programs/intent-transfer/src/bridge/processor/bridge_ntt_tokens.rs
```diff
@@ -473,6 +470,124 @@ pub struct SignedQuote {
     pub signature: [u8; 65],
 }
 
+impl SignedQuote {
```
guibescos: another idea could've been a instroducing a type:
`struct U64BE (u64)`
and manually implementing `BorshSerialize`, `BorshDeserialize` such that it uses big-endian encoding

programs/intent-transfer/src/bridge/processor/bridge_ntt_tokens.rs
```diff
@@ -473,6 +470,124 @@ pub struct SignedQuote {
...
+        bytes.extend_from_slice(&self.header.source_chain.to_be_bytes());
+        bytes.extend_from_slice(&self.header.destination_chain.to_be_bytes());
+        bytes.extend_from_slice(&self.header.expiry_time.to_be_bytes());
+        bytes.extend_from_slice(&self.base_fee.to_be_bytes());
+        bytes.extend_from_slice(&self.destination_gas_price.to_be_bytes());
+        bytes.extend_from_slice(&self.source_price.to_be_bytes());
+        bytes.extend_from_slice(&self.destination_price.to_be_bytes());
+        bytes.extend_from_slice(&self.signature);
+
+        bytes
```
guibescos: wouldn't mind a round trip test since we are manually ser/de

programs/intent-transfer/src/bridge/processor/bridge_ntt_tokens.rs
```diff
@@ -473,6 +470,124 @@ pub struct SignedQuote {
...
+        prefix.copy_from_slice(read_bytes!(4));
+
+        let mut quoter_address = [0u8; 20];
+        quoter_address.copy_from_slice(read_bytes!(20));
+
+        let mut payee_address = [0u8; 32];
+        payee_address.copy_from_slice(read_bytes!(32));
+
+        let source_chain = u16::from_be_bytes(
+            bytes[offset..offset + 2]
```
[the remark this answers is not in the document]
guibescos: you could even use array_refs! to split the buffer all at once

programs/intent-transfer/src/bridge/processor/bridge_ntt_tokens.rs
anihamde: i like this

programs/intent-transfer/src/be.rs
```diff
@@ -0,0 +1,74 @@
+use anchor_lang::{AnchorDeserialize, AnchorSerialize};
+
+#[derive(PartialEq, Debug)]
+pub struct U16BE(pub u16);
+
+impl AnchorDeserialize for U16BE {
```
guibescos: imo this module should go inside `bridge/` since it's only used by the processor there

PR #303, opened 2025-11-17
feat: paymaster bridging intent

anihamde:
This PR adds the bridging intent to the paymaster default templates. It also fixes a few issues in the Tiltfile commands.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -56,7 +56,7 @@ LOOKUP_TABLE_ADDRESSES=[
 local_resource(
     "initialize-programs",
     """
-    pnpm turbo run:initialize-chain-id -- localnet -u l -k ./tilt/keypairs/faucet.json &&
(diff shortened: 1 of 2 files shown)


Tiltfile
```diff
@@ -56,7 +56,7 @@ LOOKUP_TABLE_ADDRESSES=[
 local_resource(
     "initialize-programs",
     """
-    pnpm turbo run:initialize-chain-id -- -u l -k ./tilt/keypairs/faucet.json --chain-id localnet &&
+    pnpm turbo run:initialize-chain-id -- localnet -u l -k ./tilt/keypairs/faucet.json &&
```
guibescos: nit: can you put the position argument localnet after the `-u l -k ./tilt/keypairs/faucet.json`

guibescos (approved):

PR #305, opened 2025-11-17
fix: increase  CU limit for bridging out

cprussin:


packages/sessions-sdk-ts/src/index.ts
```diff
@@ -971,44 +973,49 @@ export const bridgeOut = async (options: SendBridgeOutOptions) => {
...
-          feeMint: options.feeConfig.mint,
-          feeSource: getAssociatedTokenAddressSync(
-            options.feeConfig.mint,
-            options.walletPublicKey,
-          ),
-        })
-        .instruction(),
-    ]),
+    [
+      ComputeBudgetProgram.setComputeUnitLimit({ units: BRIDGE_OUT_CUS }),
```
anihamde: any reason to put this outside of the definition of instructions?

PR #309, opened 2025-11-18
feat: paymaster check quote authenticity

anihamde:
This PR makes the paymaster check the authenticity of the quote provided. The paymaster now

1. checks that the signature is a valid signature of the message by a valid evm address
2. checks that the stated quoter is that which actually signed the message
3. validates that the signer matches the expected signer

### DIFF
--- a/services/paymaster/src/constraint.rs
+++ b/services/paymaster/src/constraint.rs
@@ -573,8 +573,12 @@ fn recover_signer_pubkey(signed_quote: SignedQuote) -> Result<H160, (StatusCode,
             format!("Failed to recover secp256k1 public key: {e}"),
         )
     })?;
-
-    let pubkey_hashed = keccak::hash(&secp_pubkey.serialize());
+    let secp_pubkey_serialized = secp_pubkey.serialize();
+    let pubkey_hashed = keccak::hash(
+        secp_pubkey_serialized
+            .get(1..)
+            .expect("Serialized key should be 65 bytes"),
+    );
     let evm_address = pubkey_hashed
         .0
         .get(12..)


services/paymaster/src/constraint.rs
```diff
@@ -526,8 +544,50 @@ impl DataConstraint {
     }
 }
 
+fn recover_signer_pubkey(signed_quote: SignedQuote) -> Result<H160, (StatusCode, String)> {
+    use anchor_lang::AnchorSerialize;
```
anihamde: i import these in the function because they o/w conflict with the borsh crate (which is a different version than the borsh imported by anchor 0.31.1)

services/paymaster/src/bin/tx_validator.rs
```diff
@@ -97,14 +98,27 @@ impl Network {
             Network::Testnet => "https://testnet.fogo.io",
         }
     }
+
+    fn default_ntt_quoter_address(self) -> H160 {
+        match self {
+            Network::Mainnet => [
```
anihamde: similar to other fields in the tx validator cli tool, this should not be something users of the tool have to specify. as we have it rn, there will likely be a single quoter per network mode, so users should not need to override

guibescos (commented):
For exhaustiveness, after this change you could still make the paymaster overpay by using an old quote from a time where the exchange rate was less favorable to the paymaster, but this seems to fall in the category of the "user-attacker" makes no money from it so I'm not that worried.

services/paymaster/Cargo.toml
```diff
@@ -25,6 +26,7 @@ fogo-sessions-sdk = { version = "0.7.0", path = "../../packages/sessions-sdk-rs"
 futures = {workspace = true}
 governor = {workspace = true}
 hex = {workspace = true}
+intent-transfer = {workspace = true}
```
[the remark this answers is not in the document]
anihamde: we should then publish intent-transfer. we could also just recreate the `SignedQuote` struct in fogo-paymaster, but i think that's worse practice

services/paymaster/src/constraint.rs
```diff
@@ -526,8 +544,50 @@ impl DataConstraint {
...
+    let signature = signed_quote.signature;
+    let message = &signed_quote.try_to_vec().map_err(|e| {
+        (
+            StatusCode::BAD_REQUEST,
+            format!("Failed to serialize signed quote for message hashing: {e}"),
+        )
+    })?[..100];
+
+    let r = keccak::hash(message);
+    let v = if signature[64] >= 27 {
```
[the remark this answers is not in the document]
anihamde: good call, will do

services/paymaster/src/constraint.rs
```diff
@@ -526,8 +543,67 @@ impl DataConstraint {
...
+        })?;
+
+    let secp_pubkey = libsecp256k1::recover(&msg, &sig, &recovery_id).map_err(|e| {
+        (
+            StatusCode::BAD_REQUEST,
+            format!("Failed to recover secp256k1 public key: {e}"),
+        )
+    })?;
+
+    let pubkey_hashed = keccak::hash(&secp_pubkey.serialize());
```
guibescos: do you need `&secp_pubkey.serialize()[1..65]` based on https://docs.rs/solana-secp256k1-recover/latest/src/solana_secp256k1_recover/lib.rs.html#446 ?

anihamde: goated catch!

PR #315, opened 2025-11-19
feat: hide token limits you dont have

guibescos:


packages/sessions-sdk-react/src/components/sign-in-modal.tsx
```diff
@@ -363,31 +369,64 @@ const LimitsPage = ({
...
+    case TokenDataStateType.Loaded: {
+      return (
+        <Page
+          heading="Session Limits"
+          message="Limit how many tokens this app is allowed to interact with"
+        >
+          <SessionLimits
+            enableUnlimited={enableUnlimited}
+            tokens={
+              state.type === TokenDataStateType.Error
```
guibescos: unclear what to do on error here

packages/sessions-sdk-react/src/components/sign-in-modal.tsx
```diff
@@ -363,31 +369,62 @@ const LimitsPage = ({
...
+                ? new Map(
+                    state.data.tokensInWallet.map(
+                      ({
+                        mint,
+                        amountInWallet,
+                      }: {
+                        mint: PublicKey;
+                        amountInWallet: bigint;
+                      }) => [mint, amountInWallet],
+                    ),
```
cprussin: I don't think this will work correctly because it's going to show all the tokens a user has, but due to tx size limits we will fail if a user actually tries to set limits across them.  I'd stick with `new Map()` until we figure out a way to set more than a small number of token limits.

cprussin (approved):
Overall looks really good to me but what are your thoughts about moving the filtering into `<SessionLimits>` so we don't have to apply the filters on `tokens` at each call site?

PR #321, opened 2025-11-19
feat: multiple signers

guibescos:
This implements the paymaster. The paymaster returns a random signer among those available to the caller of `sponsor_pubkey`.

In the typescript sdks I had to solve 2 issues:
- make sure `sponsor_pubkey` gets called only once so there are no two different sponsor pubkeys hanging around for a domain
- make sure `revokeSession` is called with the right sponsor

### DIFF
--- a/services/paymaster/src/api.rs
+++ b/services/paymaster/src/api.rs
@@ -154,9 +154,14 @@ impl DomainState {
                 }
 
                 Some(Box::pin(async move {
-                    self.validate_transaction_against_variation(transaction, variation, chain_index, sponsor)
-                        .await
-                        .map(|_| variation)
+                    self.validate_transaction_against_variation(
+                        transaction,
+                        variation,
+                        chain_index,
+                        sponsor,
+                    )
+                    .await
+                    .map(|_| variation)
                 }))
             })
             .collect();
@@ -366,7 +371,7 @@ async fn sponsor_and_send_handler(
             &transaction,
             &state.chain_index,
             &transaction_sponsor.pubkey(),
-            variation
+            variation,
         )
         .await
     {


packages/sessions-sdk-ts/src/connection.ts
```diff
@@ -236,9 +241,11 @@ const buildTransaction = async (
...
     await Promise.all([
       connection.rpc.getLatestBlockhash().send(),
-      connection.sponsor === undefined
-        ? getSponsor(connection, connection.sponsorCache, domain)
-        : Promise.resolve(connection.sponsor),
+      sponsorOverride
+        ? Promise.resolve(sponsorOverride)
+        : (connection.sponsor === undefined
+          ? getSponsor(connection, connection.sponsorCache, domain)
+          : Promise.resolve(connection.sponsor)),
```
cprussin:
```suggestion
      sopnsorOverride === undefined && connection.sponsor === undefined
        ? getSponsor(connection, connection.sponsorCache, domain)
        : Promise.resolve(sponsorOverride ?? connection.sponsor)),
```

minor nit but I really dislike the nested ternary

cprussin (changes requested):
The code mostly LGTM apart from some minor nits, but there's a few things I'd like to chat through:

1. What's the strategy for load balancing across the sponsors, is RNG the right way to do it?

2. How should we be selecting a sponsor in TS?  I see the code you added to be able to override the sponsor, and I see that it's only used for revoking the session with the same sponsor as the one who created the session.  That means that a given UI session will always use the same sponsor and will never load balance; maybe it makes sense to allow the UI to fetch all the sponsors and to select a different sponsor per tx rather than sticking to a single sponsor?  Would like to get your thoughts on this.

services/paymaster/src/db/config.rs
```diff
@@ -47,6 +48,7 @@ pub async fn load_config() -> Result<Config, sqlx::Error> {
             domain,
             enable_session_management,
             enable_preflight_simulation,
+            number_of_signers: NonZeroU8::new(1).expect("1 is not 0"), // TODO: Get number of signers from database
```
anihamde: can you use the `default_one` function here?

packages/sessions-sdk-ts/src/connection.ts
```diff
@@ -106,13 +106,15 @@ export const createSessionConnection = (
...
       domain: string,
+      sponsor: PublicKey | undefined,
       sessionKey: CryptoKeyPair | undefined,
       instructions: TransactionOrInstructions,
       extraConfig?: SendTransactionOptions,
     ) =>
       sendToPaymaster(
         { ...options, rpc, connection, addressLookupTableCache, sponsorCache },
         domain,
+        sponsor,
```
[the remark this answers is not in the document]
guibescos: I found a better way to do this

guibescos:
> The code mostly LGTM apart from some minor nits, but there's a few things I'd like to chat through:
> 
> 1. What's the strategy for load balancing across the sponsors, is RNG the right way to do it?
> 2. How should we be selecting a sponsor in TS?  I see the code you added to be able to override the sponsor, and I see that it's only used for revoking the session with the same sponsor as the one who created the session.  That means that a given UI session will always use the same sponsor and will never load balance; maybe it makes sense to allow the UI to fetch all the sponsors and to select a different sponsor per tx rather than sticking to a single sponsor?  Would like to get your thoughts on this.

resolved out of band

services/paymaster/src/api.rs
```diff
@@ -405,24 +429,76 @@ struct SponsorPubkeyQuery {
     #[serde(default)]
     /// Domain to request the sponsor pubkey for
     domain: Option<String>,
+    #[serde(default)]
+    #[param(value_type = String)]
+    /// Index of the sponsor to select. Can be a number or the string "autoassign" which may give you a different sponsor every time
+    index: Option<IndexSelector>,
```
anihamde: what is the use case where someone would want `autoassign`? i don't understand why anyone would want this, instead of just querying different indices manually

services/paymaster/src/api.rs
```diff
@@ -405,24 +429,76 @@ struct SponsorPubkeyQuery {
...
         enable_preflight_simulation: _,
         tx_variations: _,
+        next_autoassigned_sponsor_index,
     } = domain_state;
-    Ok(sponsor.pubkey().to_string())
+
+    let sponsor_index = if let Some(selector) = index {
+        match selector {
+            IndexSelector::Autoassign => {
+                next_autoassigned_sponsor_index.fetch_add(1, Ordering::Relaxed) % sponsors.len()
```
anihamde: i don't think this is a good feature. i would rather default to the basic pubkey (seed 0) and show different pubkeys if requested by the app. i don't get why a user would ever want this non-deterministic behavior from this GET endpoint

services/paymaster/src/constraint_templates.rs
```diff
@@ -108,18 +108,11 @@ impl InstructionConstraint {
     pub fn revoke_session_instruction_constraint() -> InstructionConstraint {
         InstructionConstraint {
             program: fogo_sessions_sdk::session::SESSION_MANAGER_ID,
-            accounts: vec![
-                AccountConstraint {
-                    index: 0,
-                    include: vec![ContextualPubkey::NonFeePayerSigner],
-                    exclude: vec![],
-                },
-                AccountConstraint {
```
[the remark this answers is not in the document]
anihamde: speaking of which we should change the `ContextualDomainKeys` to include all the sponsor keys and check all of them. otherwise, this could fail for apps that opt-in to multiple signers and continue to use the `Sponsor` account check

PR #323, opened 2025-11-20
feat: add tx variation query param support to sdk

cprussin:
This commit adds support for setting the tx variation query param.  This cannot be merged before the paymaster is upgraded to support the param because we have strict query param parsing in the paymaster, so the paymaster returns a 400 if the param is passed before the upgrade.


guibescos (approved):
lgtm once CI passes

PR #325, opened 2025-11-20
feat: tx variation param

anihamde:
This PR adds an optional param to the send_and_sponsor post request that allows specification of the variation name to validate against. This was designed in such a way to avoid config format changes or any other breaking changes.

TODOS
- [x] fix typescript SDK to specify variation params
- [x] check that spans appropriately label the variation when name is specified (even when it's wrong)
- [x] check that error propagation takes place when name is specified


guibescos (approved):
lgtm

PR #327, opened 2025-11-21
fix: allow scrolling modals

cprussin:
Our modals are usually pretty small and in nearly all cases will fit into the height of any realistic screen.  However, the new disclaimer modal is quite tall, and on sufficiently small screens (i.e. small mobile devices), the bottom of the modal gets cut off.

This commit adds some minor css tweaks to ensure the UX for modals is reasonable at those sizes and that the modal can be scrolled.

### DIFF
--- a/packages/sessions-sdk-react/src/components/modal-dialog.module.scss
+++ b/packages/sessions-sdk-react/src/components/modal-dialog.module.scss
@@ -5,17 +5,19 @@
     position: fixed;
     top: 0;
     left: 0;
-    width: 100vw;
-    height: var(--visual-viewport-height);
+    height: 100dvh;
+    width: 100dvw;
     background: theme.color("modal-overlay");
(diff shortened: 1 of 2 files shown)


packages/sessions-sdk-react/src/components/modal-dialog.module.scss
```diff
@@ -5,25 +5,27 @@
...
+    height: 100dvh;
+    width: 100dvw;
     background: theme.color("modal-overlay");
     backdrop-filter: blur(8px);
     display: grid;
     justify-content: center;
     align-items: start;
-    padding-top: theme.spacing(40);
     z-index: theme.layer("modal-dialog");
+    overflow: auto;
```
benduran: recommend also putting `-webkit-overflow-scrolling: touch` to ensure smooth scrolling on mobile devices (the inertial scrolling behavior you get in native apps)

PR #329, opened 2025-11-21
feat: wallet notional values (scaffolding)

anihamde:
This PR adds scaffolding to query the wallet's tokens' prices and show the notional values in each of the token accounts.

TODOs:

- [x] add notionals elsewhere (e.g. intent transfer page, bridging, etc.)
- [ ] uncomment the getPrice code once we add the endpoint to the web server
- [ ] test outcomes when no valid response from server

### DIFF
--- a/packages/sessions-sdk-react/src/components/notional-amount.tsx
+++ b/packages/sessions-sdk-react/src/components/notional-amount.tsx
@@ -1,5 +1,5 @@
-import { useMemo } from "react";
 import * as dnum from "dnum";
+import { useMemo } from "react";
 
 import { calculateNotional } from "../calculate-notional.js";
 
--- a/packages/sessions-sdk-react/src/components/send-token-page.tsx
+++ b/packages/sessions-sdk-react/src/components/send-token-page.tsx
@@ -317,12 +317,12 @@ const SendTokenPageImpl = ({
 
   const notionalAmount = useMemo(() => {
     if (props.isLoading || !props.amount) {
-      return undefined;
+      return;
     }
     try {
       return stringToAmount(props.amount, decimals);
     } catch {
-      return undefined;
+      return;
     }
   }, [props.isLoading, props.isLoading ? undefined : props.amount, decimals]);
 
@@ -440,14 +440,16 @@ const SendTokenPageImpl = ({
             value: props.amount,
           })}
         />
-        {!props.isLoading && props.price !== undefined && notionalAmount !== undefined && (
-          <NotionalAmount
-            amount={notionalAmount}
-            decimals={decimals}
-            price={props.price}
-            className={styles.notionalAmount}
-          />
-        )}
+        {!props.isLoading &&
+          props.price !== undefined &&
+          notionalAmount !== undefined && (
+            <NotionalAmount
+              amount={notionalAmount}
+              decimals={decimals}
+              price={props.price}
+              className={styles.notionalAmount}
+            />
+          )}
         <Button
           excludeFromTabOrder={scannerShowing}
           type="submit"
--- a/packages/sessions-sdk-react/src/components/token-list.tsx
+++ b/packages/sessions-sdk-react/src/components/token-list.tsx
@@ -93,49 +93,45 @@ export const TokenList = ({
           ]}
         >
           {(token) => {
-            if (token.isNative) {
-              return (
-                <MotionGridListItem
-                  layoutId="native-token-balance"
-                  layoutScroll
-                  textValue="Fogo"
-                  key="native-token-balance"
-                  data-is-native
-                  className={styles.token ?? ""}
-                >
-                  <div className={styles.nameAndIcon}>
-                    <img
-                      alt=""
-                      src="https://api.fogo.io/tokens/fogo.svg"
-                      className={styles.icon}
-                    />
-                    <div className={styles.nameAndMint}>
-                      <span className={styles.name}>Fogo</span>
-                      <div className={styles.mint}>NATIVE</div>
-                    </div>
+            return token.isNative ? (
+              <MotionGridListItem
+                layoutId="native-token-balance"
+                layoutScroll
+                textValue="Fogo"
+                key="native-token-balance"
+                data-is-native
+                className={styles.token ?? ""}
+              >
+                <div className={styles.nameAndIcon}>
+                  <img
+                    alt=""
+                    src="https://api.fogo.io/tokens/fogo.svg"
+                    className={styles.icon}
+                  />
+                  <div className={styles.nameAndMint}>
+                    <span className={styles.name}>Fogo</span>
+                    <div className={styles.mint}>NATIVE</div>
                   </div>
-                  <div className={styles.amountAndActions}>
-                    <div className={styles.amountAndDetails}>
-                      <span className={styles.amount}>
-                        {amountToString(token.amountInWallet, FOGO_DECIMALS)}
-                      </span>
-                    </div>
+                </div>
+                <div className={styles.amountAndActions}>
+                  <div className={styles.amountAndDetails}>
+                    <span className={styles.amount}>
+                      {amountToString(token.amountInWallet, FOGO_DECIMALS)}
+                    </span>
                   </div>
-                </MotionGridListItem>
-              );
-            } else {
-              return (
-                <TokenItem
-                  token={token}
-                  {...("onPressSend" in props && {
-                    onPressSend: props.onPressSend,
-                  })}
-                  {...("onPressToken" in props && {
-                    onPressToken: props.onPressToken,
-                  })}
-                />
-              );
-            }
+                </div>
+              </MotionGridListItem>
+            ) : (
+              <TokenItem
+                token={token}
+                {...("onPressSend" in props && {
+                  onPressSend: props.onPressSend,
+                })}
+                {...("onPressToken" in props && {
+                  onPressToken: props.onPressToken,
+                })}
+              />
+            );
           }}
         </GridList>
       );
--- a/packages/sessions-sdk-react/src/components/withdraw-page.tsx
+++ b/packages/sessions-sdk-react/src/components/withdraw-page.tsx
@@ -251,14 +251,18 @@ const WithdrawFormImpl = (
 ) => {
   const notionalAmount = useMemo(() => {
     if (props.isLoading || !props.amount) {
-      return undefined;
+      return;
     }
     try {
       return stringToAmount(props.amount, USDC.decimals);
     } catch {
-      return undefined;
+      return;
     }
-  }, [props.isLoading, props.isLoading ? undefined : props.amount, USDC.decimals]);
+  }, [
+    props.isLoading,
+    props.isLoading ? undefined : props.amount,
+    USDC.decimals,
+  ]);
 
   return (
     <Form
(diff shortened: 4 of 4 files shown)


packages/sessions-sdk-react/src/components/token-list.tsx
```diff
@@ -100,9 +104,13 @@ export const TokenList = ({
...
                 </div>
                 <div className={styles.amountAndActions}>
-                  <div className={styles.amountAndSymbol}>
-                    <span className={styles.amount}>{amountAsString}</span>
-                    {symbol && <span className={styles.symbol}>{symbol}</span>}
+                  <div className={styles.amountAndDetails}>
+                    <span className={styles.amount}>{amountAsString}{symbol && ` ${symbol}`}</span>
+                    {notionalValue !== undefined && (
+                      <span className={styles.notional}>
+                        ${notionalValue.toLocaleString(undefined, { minimumFractionDigits: 2, maximumFractionDigits: 2 })}
```
cprussin: As I'd mentioned above it would be better to use `dnum` to format so we can stick to using bigint calculations

packages/sessions-sdk-react/src/components/token-list.tsx
```diff
@@ -163,9 +171,10 @@ const LoadingToken = () => (
       </div>
     </div>
     <div className={styles.amountAndActions}>
-      <div className={styles.amountAndSymbol}>
+      <div className={styles.amountAndDetails}>
         <span className={styles.amount} />
         <span className={styles.symbol} />
```
cprussin: You need to remove this line so the loader matches the loaded state properly and you no longer have this span in the loaded state

cprussin (changes requested):
Looking like a really good start.  I don't think we're fetching prices in the right place, I think we should be adding a new hook for this and preparing to subscribe (or at least poll) for price updates.

Additionally I'd really like to see the calculations done in bigint and not integer.

One last thought is we might consider organizing the token item similar to how phantom does it:

Image

So maybe we could put the token amount & symbol below the token name, keep the notional on the right side, and maybe move the "copy mint" button to show up beside the name instead of underneath it?

packages/sessions-sdk-react/src/components/token-list.tsx
```diff
@@ -100,9 +104,13 @@ export const TokenList = ({
                   </div>
                 </div>
                 <div className={styles.amountAndActions}>
-                  <div className={styles.amountAndSymbol}>
-                    <span className={styles.amount}>{amountAsString}</span>
-                    {symbol && <span className={styles.symbol}>{symbol}</span>}
+                  <div className={styles.amountAndDetails}>
+                    <span className={styles.amount}>{amountAsString}{symbol && ` ${symbol}`}</span>
```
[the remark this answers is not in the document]
cprussin: I'm OK with this and I think I agree that the main point is the number of tokens and not the notional

packages/sessions-sdk-react/src/components/send-token-page.tsx
```diff
@@ -38,12 +41,18 @@ type Props = {
...
   symbol?: string | undefined;
   amountAvailable: bigint;
+  price?: number | undefined;
   sessionState: EstablishedSessionState;
   onPressBack: () => void;
   onSendComplete: () => void;
 };
 
 export const SendTokenPage = (props: Props) => {
+  const pricesState = usePrices([props.tokenMint.toBase58()]);
```
cprussin: change arg to a scalar not an array

packages/sessions-sdk-react/src/components/send-token-page.tsx
```diff
@@ -290,6 +300,26 @@ const SendTokenPageImpl = ({
...
+  const currentAmount = (() => {
+    if (props.isLoading || !props.amount) {
+      return 0n;
+    }
+    try {
+      return stringToAmount(props.amount, decimals);
+    } catch {
+      return 0n;
+    }
+  })();
```
cprussin: Remove IIFE

packages/sessions-sdk-react/src/components/session-panel.tsx
```diff
@@ -208,6 +208,7 @@ const Tokens = ({
               icon: token.image,
               symbol: token.symbol,
               tokenName: token.name,
+              price: token.price,
```
cprussin: Remove all changes in this file since we no longer need to thread `price` through as props

packages/sessions-sdk-react/src/components/token-list.tsx
```diff
@@ -76,9 +78,10 @@ export const TokenList = ({
             })}
         >
           {(token) => {
-            const { mint, amountInWallet, decimals, image, name, symbol } =
+            const { mint, amountInWallet, decimals, image, name, symbol: _, price } =
```
cprussin: No need to do this

packages/sessions-sdk-react/src/hooks/use-token-account-data.ts
```diff
@@ -22,11 +23,32 @@ export const useTokenAccountData = (
...
-  return useData(
+  const tokenAccountsState = useData(
     getCacheKey(network, sessionState.walletPublicKey),
     getTokenAccountData,
     {},
   );
+
+  const mints = tokenAccountsState.type === StateType.Loaded ? tokenAccountsState.data.tokensInWallet.map(t => t.mint.toBase58()) : [];
+
+  const pricesState = usePrices(mints);
```
cprussin: Remove this stuff from this component

packages/sessions-sdk-react/src/calculate-notional.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Calculate the notional (USD) value of a token amount
+ * @param amount - The token amount as a bigint
+ * @param decimals - The number of decimals for the token
+ * @param price - The price per token as a number
+ * @returns A tuple of [value, decimals] for use with dnum, or undefined if price is undefined
+ */
+export const calculateNotional = (
+  amount: bigint,
+  decimals: number,
+  price: number | undefined,
```
cprussin: I would not allow `undefined` here

packages/sessions-sdk-react/src/components/send-token-page.tsx
```diff
@@ -129,7 +137,7 @@ const useFeeTokenAccountBalance = (
     case StateType.Loaded: {
       return {
         ...accountData,
-        data:
+        amount:
```
cprussin: I actually prefer calling this `data` because that is the shape of the `State` type form `useData`, so before this change this was effectively just a way of mapping (e.g. effectively `map :: T => U => State T => State U`), but now with this change the output of `useFeeTokenAccountBalance` is no longer a `State`.

I'm not sure this matters in practice but I liked the idea that this function is returning a `State<bigint>`, and the fact that it's implemented by mapping over a different `State` isn't important to consumers.

packages/sessions-sdk-react/src/components/send-token-page.tsx
```diff
@@ -283,13 +294,24 @@ const SendTokenPageImpl = ({
...
         onChangeRecipient: (newRecipient: string) => void;
         onPressScanner: () => void;
         onChangeAmount: (newAmount: string) => void;
         feeConfig: Awaited<ReturnType<typeof getTransferFee>>;
       }
   )) => {
   const scannerShowing = !props.isLoading && props.scanner !== undefined;
+  const maxSendAmount = !props.isLoading && !props.isSubmitting && props.feeConfig.mint.equals(tokenMint)
+    ? amountAvailable - props.feeConfig.fee
+    : amountAvailable;
```
cprussin: This is fundamentally incorrect.  You have a risk that if a network connection is slow e.g. you will fall back to `amountAvailable` because of `isLoading`.  Currently the UI will prevent you from entering anything so I don't think you'll have bugs, but the code is very dangerous, because `maxSendAmount` is not valid or usable if `isLoading` or `isSubmitting` -- if some later refactor were to use `maxSendAount` in a place that _wasn't_ guarded by `isLoading`, this would lead to subtle bugs that are hard to understand and would not be caught if someone wasn't explicitly testing USDC.

I think if you want to do this, you need to think about restructuring the code a bit so you don't have `maxSendAmount` ever in a state where the value it holds is incorrect.

cprussin (changes requested):

packages/sessions-sdk-react/src/components/notional-amount.tsx
```diff
@@ -0,0 +1,49 @@
...
+}: Props) => {
+  if (validationError) {
+    return (
+      <div className={clsx(className, styles.error)}>
+        Invalid amount specified
+      </div>
+    );
+  }
+
+  if (amount.length === 0) {
```
anihamde: possibly this check belongs in the JSX where `NotionalAmount` is invoked? but in the spirit of deduplication i put this here, even though semantically i think it's better as a check to determine whether `NotionalAmount` is even invoked. open to suggestions here

packages/sessions-sdk-react/src/components/token-list.tsx
```diff
@@ -150,6 +100,81 @@ export const TokenList = ({
...
+            <TruncateKey keyValue={mint} />
+          </CopyButton>
+        </div>
+      </div>
+      <div className={styles.amountAndActions}>
+        <div className={styles.amountAndDetails}>
+          <span className={styles.amount}>{amountAsString}</span>
+          {price.type === PriceDataStateType.Loaded && (
+            <NotionalAmount
+              amount={amountAsString}
```
anihamde: i don't love this conversion to string and then back within NotionalAmount, but i preferred this to what was here before, which was essentially the internals of the NotionalAmount component. We could make the amount prop `string | bigint`?

packages/sessions-sdk-react/src/components/notional-amount.tsx
cprussin: It's a bit weird to pass `amount` as a string for sure but I wouldn't overthink it, this is a reasonable api for how it's used in practice in all current cases.  We can revisit later when we build more features and see how things transform.

packages/sessions-sdk-react/src/components/token-amount-input.tsx
```diff
@@ -12,11 +12,20 @@ export const TokenAmountInput = ({
   max,
   gt,
   lt,
+  onValidationChange,
```
cprussin: You aren't using this in the end and there was some indication that it caused render issues so we should revert this until we go to revisit down the line

packages/sessions-sdk-react/src/calculate-notional.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param price - The price per token as a number
+ * @returns The product as a Dnum
+ */
+export const calculateNotional = (
+  amount: bigint,
+  decimals: number,
+  price: number,
+): dnum.Dnum => {
+  const priceDnum = dnum.from(price);
+  const amountDnum = [amount, decimals] as dnum.Dnum;
```
cprussin: I think you can just do `dnum.from([amount, decimals])` which I always prefer over the `as` keyword (the `as` keyword can typecast in dangerous ways)

cprussin (approved):
Minor nits but pretty much lgtm

PR #333, opened 2025-11-24
chore(paymaster): add name & version to variation

alexcambose:


services/paymaster/src/bin/seed.rs
```diff
@@ -97,12 +96,36 @@ async fn insert_variation(
     domain_config_id: &Uuid,
     variation: &TransactionVariation,
 ) -> Result<Uuid, sqlx::Error> {
+    let transaction_variation_json = match variation {
+        TransactionVariation::V0(v) => {
+            // Convert Pubkeys to strings (base58 format) before serializing
+            let pubkey_strings: Vec<String> = v
```
anihamde: do you need the explicit type specified here?

services/paymaster/src/bin/seed.rs
```diff
@@ -118,12 +141,18 @@ pub async fn seed_from_config(config: &Config) -> Result<(), anyhow::Error> {
...
         tracing::info!("Seeding database from config");
         for domain in &config.domains {
-            let domain_url = Url::parse(&domain.domain).unwrap();
-            let host = domain_url
-                .host_str()
-                .ok_or(anyhow::anyhow!("Invalid URL"))?;
-            let user = insert_user(host).await?;
-            let app = insert_app(&user, host).await?;
+            // exception for "sessions" domain, insert that as url in the db
+            let host = if domain.domain == "sessions" {
```
anihamde: i don't think we should hardcode this, maybe we should try url parsing and fallback to using the raw domain name if url parsing fails?

services/paymaster/src/db/config.rs
```diff
@@ -1,32 +1,46 @@
...
+/// Database row where JSON is untyped.
 #[derive(FromRow)]
-struct Row {
+struct RawRow {
     domain_id: Uuid,
     domain: String,
     enable_session_management: bool,
     enable_preflight_simulation: bool,
-    transaction_variation: Json<TransactionVariation>,
+    version: Option<String>,
```
anihamde: why are these options? we should make them nonnull in the db, no?

services/paymaster/src/db/config.rs
```diff
@@ -35,26 +49,102 @@ pub async fn load_config() -> Result<Config, sqlx::Error> {
...
-            transaction_variation.0.name().to_string(),
-            transaction_variation.0,
-        );
+        let transaction_variation_fin: TransactionVariation;
+
+        if let (Some(version), Some(name), Some(transaction_variation)) =
+            (version, name, transaction_variation)
+        {
+            if version == "v0" {
+                // Extract the whitelisted_programs array as Vec<String>
```
anihamde: same as above

services/paymaster/migrations/20251020133349_init.down.sql
```diff
@@ -4,5 +4,6 @@ DROP TABLE IF EXISTS variation;
 DROP TABLE IF EXISTS domain_config;
 DROP TABLE IF EXISTS app;
 DROP TABLE IF EXISTS "user";
+DROP TYPE IF EXISTS variation_version;
```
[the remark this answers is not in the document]
alexcambose: Can you explain why do you prefer multiple migration files? 
I don't really see why we would want to split this up into multiple files for the initial migration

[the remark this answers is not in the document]
anihamde: similar to smaller PRs, smaller migrations means more scoped changes, easier migrations up and down, less room for error. general practice, i'm not too picky about this case

services/paymaster/src/db/config.rs
anihamde: https://github.com/fogo-foundation/fogo-sessions/pull/333/files/4968ade4b131d4c0ed620fb8bae8b2466787bbe7#r2561291750

services/paymaster/src/db/config.rs
```diff
@@ -27,7 +27,51 @@ struct RawRow {
...
+            .filter_map(|v| v.as_str().map(|s| s.to_string()))
+            .collect(),
+        _ => {
+            return Err(anyhow::anyhow!(
+                "Missing or invalid whitelisted_programs field"
+            ));
+        }
+    };
+
+    // Convert Vec<String> to Vec<Pubkey>
```
anihamde: can get rid of these comments, they're not helpful for understanding the code. generally i only recommend including inline comments when they disclose some information a new reader couldn't easily attain by just reading the code. otherwise, the comment is unhelpful and could be a lightning rod for becoming confusing/outdated

services/paymaster/src/db/config.rs
```diff
@@ -70,74 +114,26 @@ pub async fn load_config() -> Result<Config, sqlx::Error> {
...
-
-                transaction_variation_fin =
-                    TransactionVariation::V1(VariationOrderedInstructionConstraints {
-                        name,
-                        instructions,
-                        max_gas_spend: max_gas_spend.map(|v| v as u64).unwrap_or(0),
-                    });
-            }
-
+                    max_gas_spend as u64,
```
anihamde: casting via `as` is not the best practice because the conversion isn't checked. e.g. for the `i64 -> u64` conversion here, the `as` keyword will wrap negative `i64`s into large positive `u64`s. instead i would try `u64::try_from(max_gas_spend)?` with appropriate error mapping and propagation

services/paymaster/src/db/config.rs
```diff
@@ -14,17 +14,21 @@ use uuid::Uuid;
 
 use crate::db::pool;
 
-/// Database row where JSON is untyped.
 #[derive(FromRow)]
-struct RawRow {
-    domain_id: Uuid,
+struct DomainConfigRow {
```
anihamde: nit: you don't need to suffix it `Row`. that naming ties it to the name of the table. maybe `DomainConfigMetadata`? i think there's a better way to name this but don't have a suggestion off the top of my head

services/paymaster/src/db/config.rs
```diff
@@ -14,17 +14,21 @@ use uuid::Uuid;
...
     enable_session_management: bool,
     enable_preflight_simulation: bool,
-    version: Option<String>,
-    name: Option<String>,
-    max_gas_spend: Option<i64>,
-    transaction_variation: Option<Json<Value>>,
+}
+
+#[derive(FromRow)]
+struct VariationRow {
```
anihamde: lo mismo aquí

anihamde: i think better to just call this `Variation` or `DomainVariation`

services/paymaster/src/db/config.rs
```diff
@@ -139,11 +155,14 @@ pub async fn load_config() -> Result<Config, anyhow::Error> {
                 transaction_variation_fin,
             );
         } else {
-            continue;
+            tracing::warn!(
```
anihamde: nice!

anihamde (commented):
almost there. let's get into the practice of running `cargo sqlx prepare` whenever we have PRs that change the sqlx queries

services/paymaster/src/db/config.rs
```diff
@@ -1,63 +1,186 @@
...
+    for Variation {
+        domain_config_id,
+        version,
+        name,
+        max_gas_spend,
+        transaction_variation,
+    } in variation_rows
+    {
+        if let Some(domain_ref) = domain_map.get_mut(&domain_config_id) {
+            let transaction_variation_fin: TransactionVariation = match version.as_str() {
```
anihamde: you don't need the explicit type spec here

anihamde (approved):

services/paymaster/src/db/config.rs
alexcambose: 👍 I'll change this in the other PRs

PR #335, opened 2025-11-24
chore(paymaster): remove app_user & switch to uuidv4

alexcambose:
* Removed app_user
* Switched to uuid v4


services/paymaster/migrations/20251020133349_init.up.sql
```diff
@@ -1,4 +1,3 @@
-CREATE TYPE app_role AS ENUM ('owner', 'admin');
```
anihamde: it seems we didn't have some of these in the down migration. we should be careful about making sure `down` composed with `up` leads back to the original state, i.e.

$$ \text{down} \circ \text{up} = \text{id}$$

anihamde: we should generally test this as part of the vetting process for a PR. let's do this before we merge this PR, to make sure that the down and up migrations are truly inverses.

.prettierignore
```diff
@@ -17,3 +17,4 @@ packages/
 scripts/
 pnpm-lock.yaml
 tilt/keypairs/
+.sqlx/
```
anihamde: i actually want this in the git history. i want to see the query files

PR #346, opened 2025-11-25
Add update metadata script

cprussin:


scripts/ts/src/update-metadata.ts
```diff
@@ -0,0 +1,82 @@
...
+      description: "Mint to set the metadata for",
+      demandOption: true,
+      coerce: (mint: string) => new PublicKey(mint),
+    })
+    .parse();
+
+  const provider = createAnchorProvider(args);
+
+  const umi = createUmi(provider.connection.rpcEndpoint);
+  if (provider.wallet.payer === undefined) {
```
guibescos: this won't work with ledger unlike the other scripts here, maybe leave a comment explaining that

scripts/ts/src/update-metadata.ts
```diff
@@ -0,0 +1,82 @@
...
+        type: "string",
+        description: "Token name",
+      },
+      symbol: {
+        type: "string",
+        description: "Token symbol",
+      },
+      tokenUri: {
+        type: "string",
+        description: "Token metadata uri",
```
anihamde: worthwhile to make it explicit that these are the values to be updated _to_?

anihamde (approved):

PR #348, opened 2025-11-25
fix: gracefully handle transferring in with no Sol on solana

cprussin:


packages/sessions-sdk-react/src/components/deposit-page.tsx
```diff
@@ -21,71 +21,127 @@ import { StateType, useData } from "../hooks/use-data.js";
...
+            headline="No SOL in your Solana wallet"
+            error="You must have SOL in your Solana wallet to pay gas on Solana to transfer to Fogo."
+          />
+        );
+      } else if (solanaBalances.data.usdc.amount === "0") {
+        return (
+          <FetchError
+            className={styles.fetchError}
+            headline="No USDC in your Solana wallet"
+            error="You have no USDC on Solana to transfer to Fogo."
```
guibescos: should this one be first?
like if the wallet is completely empty seems like the most logical error to raise is "you don't have USDC"

guibescos (approved):

anihamde (approved):

PR #351, opened 2025-11-25
fix: paymaster admin db schema

anihamde:
This PR fixes the db schema incompatibilities with the paymaster admin UI. It also makes some minor changes to some layouts.

### DIFF
--- a/apps/paymaster-admin/src/db-schema.ts
+++ b/apps/paymaster-admin/src/db-schema.ts
@@ -21,6 +21,7 @@ const TimeStr = z.preprocess((val) => {
 export const PrimitiveDataValueSchema = z.union([
   z.object({ U8: z.number().int().min(0).max(255) }),
   z.object({ U16: z.number().int().min(0).max(65_535) }),
+  // prettier-ignore
   z.object({ U32: z.number().int().min(0).max(0xFF_FF_FF_FF) }),
   z.object({ U64: u64 }),
   z.object({ Bool: z.boolean() }),


apps/paymaster-admin/src/server/paymaster.ts
```diff
@@ -75,11 +79,25 @@ export const fetchUserPaymasterData = async (walletAddress: string) => {
...
+    console.log("User not found in database:", walletAddress);
+    return {
+      id: null,
+      username: null,
+      wallet_address: walletAddress,
+      created_at: null,
+      updated_at: null,
+      apps: [],
+    };
+  }
```
alexcambose: We shouldn't return this. This function should either return an object with the `UserSchama` structure or potentially `undefined`.
I would suggest deleting and just keeping the code below or if you want to make it very explicit also adding  `if (!rows[0]) { return undefined; }`

apps/paymaster-admin/src/server/paymaster.ts
```diff
@@ -66,7 +69,8 @@ export const fetchUserPaymasterData = async (walletAddress: string) => {
               ) AS dc_row
             ) AS domain_configs
           FROM app a
-          WHERE a.user_id = u.id
+          INNER JOIN app_user au ON au.app_id = a.id
+          WHERE au.user_id = u.id
```
alexcambose: fyi I deleted the `app_user` table here https://github.com/fogo-foundation/fogo-sessions/pull/335, but we can merge this and update this query later

apps/paymaster-admin/src/server/paymaster.ts
alexcambose: nice, in the future we might want to handle this edge case right after the user connects their wallet and display a message/error page, so that we won't need to handle this on every page that uses the user data.

alexcambose (approved):

PR #353, opened 2025-11-26
feat: paymaster-sync-config

alexcambose:
This PR introduces changes the seeding script to also update existing domains and variations from a `config.toml` file, the functionality is as follows now: 
* insert users
* insert apps
* insert or update existing domains by domain name
* insert or update existing variations by variation name

Initially running this command against an empty db will populate the DB with new items, and running this after config file changes will trigger updates.
```
cargo run --bin paymaster-config-sync -- --config="./tilt/configs/paymaster.toml" --db-url="postgres://paymaster:paymaster@localhost:5432/paymaster"
```

**note that in case of deleting from the config file, the changes will not be reflected in the DB*


services/paymaster/src/bin/config_sync.rs
```diff
@@ -71,7 +82,46 @@ async fn insert_app(user_id: &Uuid, name: &str) -> Result<Uuid, sqlx::Error> {
...
+        "SELECT id FROM domain_config WHERE app_id = $1 AND domain = $2",
+    )
+    .bind(app_id)
+    .bind(domain.domain.to_string())
+    .fetch_optional(pool())
+    .await?;
+
+    if let Some(domain_config_result) = existing_domain_config {
+        let id = domain_config_result.0;
+        let res = sqlx::query(
```
anihamde: i think we should run `cargo sqlx prepare` whenever you update queries so that sql metadata gets updated and queries get validated. this prevents any schema errors in the queries at runtime

anihamde: we should use `query!` and/or `query_as!` to get the compile-time checking

services/paymaster/src/bin/config_sync.rs
```diff
@@ -0,0 +1,249 @@
...
+
+pub fn load_file_config(config_path: &str) -> Result<Config> {
+    let config: Config = config::Config::builder()
+        .add_source(File::with_name(config_path))
+        .build()?
+        .try_deserialize()?;
+    Ok(config)
+}
+
+async fn run_seed(db_url: &str, config_path: &str) -> anyhow::Result<()> {
```
anihamde: nit, this is no longer called seed right?

services/paymaster/src/bin/config_sync.rs
```diff
@@ -0,0 +1,249 @@
...
+        // try to parse the domain as a url and get the host or just return the domain if it's not a valid url
+        let host = match Url::parse(&domain.domain) {
+            Ok(url) => url
+                .host_str()
+                .map(|h| h.to_string())
+                .unwrap_or_else(|| domain.domain.clone()),
+            Err(_) => domain.domain.clone(),
+        };
+
+        let user = insert_user(&host).await?;
```
anihamde: nit: maybe name the functions something like `insert_user_if_new`, etc.?

services/paymaster/migrations/20251020133349_init.up.sql
```diff
@@ -35,5 +39,7 @@ CREATE TABLE variation (
   max_gas_spend bigint,
   transaction_variation jsonb NOT NULL,
   created_at timestamptz NOT NULL DEFAULT now(),
-  updated_at timestamptz NOT NULL DEFAULT now()
+  updated_at timestamptz NOT NULL DEFAULT now(),
+
+  CONSTRAINT variation_domain_name_unique UNIQUE (domain_config_id, name)
```
anihamde: this one makes sense to me

PR #355, opened 2025-11-27
fix: allow dot in token metadata symbols

guibescos:
This is preventing token limits with USDC.s


programs/session-manager/src/message.rs
```diff
@@ -131,6 +132,16 @@ impl UiTokenAmount {
     }
 }
 
+fn symbol_or_mint<I, E>(input: I) -> IResult<I, I, E>
+where
+    I: Input,
+    E: ParseError<I>,
+    <I as Input>::Item: AsChar,
+{
+    take_while1(|c: <I as Input>::Item| c.is_alphanum() || ['.'].contains(&c.as_char()))
```
guibescos: could also allow more characters?

cprussin (approved):
Nice catch

PR #359, opened 2025-11-28
chore: remove intent transfer template

guibescos:
Remove intent transfer templates now that transfers only happen through the internal domain

### DIFF
--- a/services/paymaster/src/config_manager/config.rs
+++ b/services/paymaster/src/config_manager/config.rs
@@ -76,7 +76,7 @@ where
 pub struct Config {
     pub domains: Vec<Domain>,
 }
-pub const DEFAULT_TEMPLATE_MAX_GAS_SPEND: u64 = 100_000;
+pub const DEFAULT_TEMPLATE_MAX_GAS_SPEND: u64 = 15_000;
 
 impl Config {
     /// Populate default tx variations for each domain.


tilt/configs/paymaster.toml
```diff
@@ -70,4 +70,144 @@ required = true
...
+# instruction = 1 (BridgeNttTokens)
+[[domains.tx_variations.instructions.data]]
+start_byte = 0
+data_type = "U8"
+constraint = { EqualTo = [ {U8 = 1} ] }
+
+[[domains.tx_variations.instructions.data]]
+start_byte = 1
+data_type = "NttSignedQuote"
+constraint = { EqualTo = [ {NttSignedQuoter = "0x5241c9276698439fef2780dbab76fec90b633fbd"} ] }
```
guibescos: I checked this config with the transaction validator tool

services/paymaster/src/config_manager/config.rs
```diff
@@ -77,12 +76,12 @@ where
 pub struct Config {
     pub domains: Vec<Domain>,
 }
-pub const DEFAULT_TEMPLATE_MAX_GAS_SPEND: u64 = 100_000;
+pub const DEFAULT_TEMPLATE_MAX_GAS_SPEND: u64 = 15_000;
```
guibescos: drive-by: 15_000 as priority fees are not needed yet on Fogo

PR #361, opened 2025-11-30
feat: show native balance; add enable native transfers w/ wallet

cprussin:
This change shows native token balance in the session widget and adds a stopgap mechanism for enabling native transfers, by allowing transactions to be signed with the wallet instead of the session.

The UX here has some downsides:

1. The wallet will show a big scary warning message about the transaction being untrusted
2. The wallet will show SOL instead of FOGO as the token being transferred
3. Currently I haven't added the necessary changes to enable sending tokens using this new mechanism (but this is something I can do in a follow-up if we decide to move forward on this)

However, despite these downsides, this will allow users to interact with native tokens even before sessions can actually support them, so we might consider this as a temporary stopgap solution.

### DIFF
--- a/.changeset/nice-trains-clap.md
+++ b/.changeset/nice-trains-clap.md
@@ -1,0 +1,6 @@
+---
+"@fogo/sessions-sdk-react": patch
+"@fogo/sessions-sdk": patch
+---
+
+Show native token balance & allow signing transactions with wallet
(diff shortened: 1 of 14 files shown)


cprussin:
Closing this because we merged the native balance feature in a separate PR and we don't plan to move forward with the wallet signing feature unless absolutely needed

PR #363, opened 2025-12-01
feat: component library init

alexcambose:
This PR just sets up the component library storybook package with a dummy Link component.
Note there's no linking with the react sdk or other projects.

Most of the changes here are copied from pyth-crosschain, but upgraded storybook to v10.

### DIFF
--- a/packages/component-library/package.json
+++ b/packages/component-library/package.json
@@ -10,14 +10,12 @@
     "./Link": {
       "default": "./dist/Link/index.mjs"
     },
-    "./package.json": "./package.json",
     "./theme": "./dist/theme.scss"
   },
   "scripts": {
     "build": "ts-duality --noCjs",
     "build:scss": "copyfiles -u 1 \"src/**/*.scss\" ./dist",
     "build:storybook": "storybook build",
-    "build:svg": "copyfiles -u 1 \"src/**/*.svg\" ./dist",
     "clean": "rm -rf ./dist",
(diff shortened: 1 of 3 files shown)


packages/component-library/package.json
```diff
@@ -0,0 +1,80 @@
...
+  "exports": {
+    "./Link": {
+      "default": "./dist/Link/index.mjs"
+    },
+    "./package.json": "./package.json",
+    "./theme": "./dist/theme.scss"
+  },
+  "scripts": {
+    "build": "ts-duality --noCjs",
+    "build:scss": "copyfiles -u 1 \"src/**/*.scss\" ./dist",
```
cprussin: I think we should be distributing actually compiled css right?  Any reason why we want to distribute untranspiled scss?

packages/component-library/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "build:storybook": "storybook build",
+    "build:svg": "copyfiles -u 1 \"src/**/*.svg\" ./dist",
```
cprussin: I think you're safe to omit this, we don't have svgs (there's so few custom svgs here that for the cases we do have -- basically just the logo -- I just made them react components instead of using svgr)

packages/component-library/package.json
alexcambose: yeah I didn't really give much thought into the build system here, I have another PR (still in progress) that addresses all the build stuff

PR #367, opened 2025-12-02
split index.ts into individual files, no code changes

nonergodic:
first commit of SDK refactor [summary PR](https://github.com/fogo-foundation/fogo-sessions/pull/357)


packages/sessions-sdk-ts/src/index.ts
```diff
@@ -1,70 +1,7 @@
...
-import { SESSIONS_INTERNAL_PAYMASTER_DOMAIN } from "./context.js";
-import {
-  importKey,
-  signMessageWithKey,
-  verifyMessageWithKey,
-} from "./crypto.js";
+export * from "./session.js";
+export * from "./login-token.js";
+export * from "./transfer.js";
+export * from "./bridge.js";
```
cprussin: Let's not use barrel exports, they can be unperformant for consumers, depending on the environment (e.g. https://www.atlassian.com/blog/atlassian-engineering/faster-builds-when-removing-barrel-files) and are very hard to trace (reading a package like @solana/kit that makes heavy use of barrel exports and trying to figure out where things come from is unnecessarily painful for no really good reason).

I think we should reorganize the code, but the code is not functionally reorganized in any significant way if it's all spread and re-exported from the same place.  Instead, we should reorganize into logically delineated subpath exports.  This would solve the very practical problem of allowing imports to pull in only dependencies that are really necessary for whatever functionality is getting called -- in particular libraries like the Wormhole SDKs are not transpiled correctly and do not work on Node (the ESM that is shipped in those libraries breaks on node because it isn't transpiled correctly to follow Node ESM requirements), which causes issues with apps that use SSR and also try to statically import the Sessions SDK.  Reorganizing into delineated subpath exports would clean up the code in a logical way that would also resolve this issue and would minimize bundle size.

cprussin (changes requested):
I love the reorganization, but I think we need to think more logically about how we reorganize.  In particular, reorganizing just to re-spread everything into the top level isn't functionally much of a reorganization, and there's a very good use case for introducing some subpath exports here to help reduce bundle size and resolve some SSR issues.  I think we should:

1. Determine which subpath exports are logical given the logical grouping of functionality that would be imported together (e.g. have a subpath for session management, one for bridging, one for transfers, etc)

2. Reorganize the code into those subpath exports, and

3. Avoid barrel exports

nonergodic:
> In short, I think what you propose makes perfect sense, except that I don't see why you'd do it now, rather than after.

Having thought about it a bit more, I assume the difference here comes down to expectations of how long merging the entire thing "should" take. If you expect it to take a while, it makes sense to make small changes for outsized gains now (which seems to be the case for what you propose, if it indeed helps with reducing bundle size (I have no idea and I have to wonder how shitty bundlers are if they can't tell that they can get rid of entire subtrees when you obviously don't pull anything from that part, e.g. if you only use the session management parts without any of the intent transfer parts)).

PR #373, opened 2025-12-03
feat: paymaster wallet authentication

alexcambose:
Very simple wallet authentication, things to note:
* ~~We store the signed session token inside cookies so we can have ssr with user-related data~~
* ~~No db interaction in this PR but authentication will include that too~~

* Everything UI related here will be changed
* Db interaction to fetch all user data and display an error state if the connected wallet is not registered as a paymaster user
* SPA managed via react context and nextjs api routes

https://github.com/user-attachments/assets/48f5ba6b-95e3-40af-b38e-4dab22a611cb

### DIFF
--- a/apps/paymaster-admin/src/components/Root/layout.tsx
+++ b/apps/paymaster-admin/src/components/Root/layout.tsx
@@ -2,7 +2,7 @@
 import { SessionStateType, useSession } from "@fogo/sessions-sdk-react";
 
 import { Auth } from "../Auth";
-import { Navbar } from "../NavBar";
+import { Navbar } from "../Navbar";
 
 export const AuthenticationLayout = ({
   children,


apps/paymaster-admin/src/components/Home/index.tsx
```diff
@@ -0,0 +1,11 @@
+"use client";
```
cprussin: Same comment

cprussin (changes requested):
Mostly minor things, biggest thing is that I don't think there's a good reason to deal with setting the session token to the cookie server side.

More fundamentally I'm not convinced that SSR-ing the authenticated app UI code is worth the effort.  It's kind of nasty to sync to the cookie, and in any case SSR code that relies on cookies cannot be cached and won't be used in SEO so the benefit of SSR is just marginally faster boot times (basically the initial render being html instead of a loading state while json fetches & the subsequent route renders).  Arguably it will actually be faster to not do SSR on the authenticated pages unless you're very, very careful with it, since you can cache all the static stuff and load the loading state very fast, but you can't cache all the stuff that relies on auth state, so unless you're very diligent about putting in nextjs loading pages, you'll actually make performance worse.

I'm not sure that it will be worth the added cost on you to do any of it server-side, I would consider just having the app shell SSR and having all the authenticated client state be driven by good old fashioned client-side SPA magic.

apps/paymaster-admin/src/app/api/user-data/route.ts
```diff
@@ -0,0 +1,35 @@
...
+  cacheLife("seconds");
+  return fetchUserPaymasterData(walletAddress);
+};
+
+export const GET = async (request: NextRequest) => {
+  const auth = request.headers.get("authorization");
+  const sessionToken = auth?.startsWith("Bearer ") ? auth.slice(7) : undefined;
+
+  if (!sessionToken)
+    return NextResponse.json({ error: "Missing token" }, { status: 401 });
```
cprussin: Small nit but no reason to use json for errors.  Also please use braces around clauses.

apps/paymaster-admin/src/components/Home/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { SessionButton } from "@fogo/sessions-sdk-react";
+
+import { useUserData } from "../user-data-context";
+
+export const Home = () => {
+  const { userData, isLoading, error, isUserNotFound, refetch } = useUserData();
```
cprussin: We should implement `useUserData` in terms of `useData` from https://github.com/fogo-foundation/fogo-sessions/blob/main/packages/sessions-sdk-react/src/hooks/use-data.ts (feel free to move that to the component library so it's reusable) -- as it stands you are relying on `useUserData` to never return mutually exclusive states or you will have a broken page.  In particular, a very common example -- what if you moved to a stale-while-revalidate fetch model (which you should do and I will comment that below) and thus it's possible to both have `userData` and also have `isLoading` be true?  Your UI would be broken and the type structure you have in place would not prevent it.

You are currently relying on runtime data assumptions which are not reflected in the type system to guarantee the UI is actually correct, and you can instead use the type system to verify and guarantee this.  In general, you should use the type system to explicitly and clearly designate states which are mutually exclusive, so that you can guarantee at compile time that the consumer is correct.

apps/paymaster-admin/src/components/user-data-context.tsx
```diff
@@ -0,0 +1,99 @@
...
+  userData: User | undefined;
+  isLoading: boolean;
+  error: Error | undefined;
+  isUserNotFound: boolean;
+  refetch: () => Promise<void>;
+};
+
+const UserDataContext = createContext<UserDataContextValue | undefined>(
+  undefined,
+);
```
cprussin: I don't think I agree with the idea of putting all this data into context:

1. I don't think it's appropriate to use context unless you really need to for truly global state.  This isn't global state, and it's not hard to thread the data down into the components that need it.  I don't think the cost of using a context (losing the obvious flow of data through the application) is justified by the benefit (less prop drilling) in this case.
2. I don't think we should be using a single endpoint for all data, especially as we add CRUD to update things like specific configs for specific apps, it seems unlikely that a single endpoint will remain the long term model here
3. It's much easier to turn something into a context later if we decide the prop drilling is getting unweildy than it is to go the other way around.  I'd prefer to start  by not using context unless we really absolutely must.

apps/paymaster-admin/src/config/server.ts
```diff
@@ -0,0 +1,53 @@
...
+  : (key: string) => process.env[key];
+
+export const GOOGLE_ANALYTICS_ID = defaultInProduction(
+  "GOOGLE_ANALYTICS_ID",
+  "G-8SYVHSS051",
+);
+
+export const RPC = getEnvOrDefault("RPC", "https://testnet.fogo.io");
+
+// TODO Replace this domain with https://paymaster-admin.fogo.io or something after setting up the domain in the paymaster
```
cprussin: We should just do this now before we merge this, it's very simple nowadays

apps/paymaster-admin/src/db-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+  .max(44);
+const u16 = z.number().int().min(0).max(65_535);
+const u64 = z.number().int().min(0).max(Number.MAX_SAFE_INTEGER);
+
+const TimeStr = z.preprocess((val) => {
+  if (val instanceof Date) return val;
+  if (typeof val === "string" || typeof val === "number") {
+    return new Date(val);
+  }
+  return val; // will prob fail against z.date() below
```
cprussin: I particularly dislike the early return in cases like these, this is IMO way more of a pain to read than reading control flow because when I look at line 17, I actually cannot trust that this is the correct return without reading the rest of the code, which may or may not be relevant to the conditions guarding this arm.  In other words, I cannot make any guarantees about this line without reading the whole function and understanding how every early return works first, but it's also not obvious that this is the case without reading the function top-to-bottom.

Not as big of an issue with a 5 line function, but if the function were to grow significantly, that can be really hard  to predict.

apps/paymaster-admin/src/db-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+export const UserSchema = z.object({
+  id: UUID,
+  username: z.string(),
+  wallet_address: z.string(),
+  created_at: TimeStr,
+  updated_at: TimeStr,
+  apps: z.array(AppWithDomainConfigsSchema),
+});
+
+export type User = z.infer<typeof UserSchema>;
```
cprussin: This file is awesome, very nicely done on the zod schemas

apps/paymaster-admin/package.json
```diff
@@ -10,19 +14,27 @@
     "start:prod": "next start --port 3002",
     "test:format": "prettier --check .",
     "test:lint": "eslint . --max-warnings 0",
-    "test:types": "tsc"
+    "test:types": "tsc",
+    "pull:env": "[ $CI ] || VERCEL_ORG_ID=team_M8yTQcRiKOduAwHNMhHQInWq VERCEL_PROJECT_ID=prj_0YVLB5TIHu7VXdMyib1Zmjq4iwtZ vercel env pull"
```
cprussin: Minor nit but let's sort alphabetically please

apps/paymaster-admin/src/config/server.ts
```diff
@@ -0,0 +1,53 @@
...
+const defaultInProduction = IS_PRODUCTION_SERVER
+  ? getEnvOrDefault
+  : (key: string) => process.env[key];
+
+export const GOOGLE_ANALYTICS_ID = defaultInProduction(
+  "GOOGLE_ANALYTICS_ID",
+  "G-8SYVHSS051",
+);
+
+export const RPC = getEnvOrDefault("RPC", "https://testnet.fogo.io");
```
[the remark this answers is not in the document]
alexcambose: hmm, I'm not sure about this. Mainnet and testnet dbs will be different and might be messy to have both on the same app..

I need to think about this.

apps/paymaster-admin/src/db-schema.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export const PrimitiveDataTypeSchema = z.union([
+  z.literal("U8"),
+  z.literal("U16"),
+  z.literal("U32"),
+  z.literal("U64"),
+  z.literal("Bool"),
+  z.literal("Pubkey"),
+  z.object({ Bytes: z.object({ length: z.number().int().min(0) }) }),
+]);
```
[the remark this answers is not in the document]
alexcambose: nice

apps/paymaster-admin/src/app/api/auth/user-data/route.ts
```diff
@@ -0,0 +1,27 @@
...
+  "use cache";
+  cacheTag("user-data");
+  cacheLife("seconds");
+  return fetchUserPaymasterData(walletAddress);
+};
+
+export const GET = async (request: NextRequest) => {
+  const walletAddress = request.headers.get("x-authenticated-user");
+  if (!walletAddress) {
+    return NextResponse.json("Unauthorized", { status: 401 });
```
cprussin: While a string is technically valid json, you should construct text responses with `new Response(body, { status: 401 })`.

apps/paymaster-admin/src/client/paymaster.ts
```diff
@@ -0,0 +1,50 @@
...
+  url.searchParams.set("sessionToken", sessionToken);
+  const response = await fetch(url, {
+    headers: { Authorization: `Bearer ${sessionToken}` },
+  });
+  if (response.status === 404) {
+    throw new UserNotFoundError();
+  } else if (response.ok) {
+    return UserSchema.parse(await response.json());
+  } else {
+    throw new Error("Failed to fetch user data");
```
cprussin: Maybe add failure details here so we can actually debug this if it were to be reported

apps/paymaster-admin/src/components/Home/index.tsx
```diff
@@ -0,0 +1,19 @@
+"use client";
+import { StateType } from "@fogo/component-library/useData";
+import { SessionButton } from "@fogo/sessions-sdk-react";
+
+import { useUserData } from "../../client/paymaster";
+
+export const Home = () => {
+  const userData = useUserData();
```
cprussin: If you follow my advice above, generally I'd restructure this component to switch on session state.  If the session state is established, then pass `sessionState` to a child component that then calls `useUserData`.

apps/paymaster-admin/src/client/paymaster.ts
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export const fetchUserData = async (sessionToken: string) => {
+  const url = new URL("/api/auth/user-data", globalThis.location.origin);
+  url.searchParams.set("sessionToken", sessionToken);
+  const response = await fetch(url, {
+    headers: { Authorization: `Bearer ${sessionToken}` },
+  });
+  if (response.status === 404) {
+    throw new UserNotFoundError();
```
cprussin: I don't like the idea of throwing exceptions for things that are perfectly recoverable, IMO we should return a union type for the possible states here and handle it as a normal value.  Exceptions should be reserved for things that are not recoverable, see https://wiki.c2.com/?DontUseExceptionsForFlowControl

PR #378, opened 2025-12-04
chore: use new paymaster domains

guibescos:


services/paymaster/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fogo-paymaster"
-version = "0.5.0"
+version = "0.5.1"
```
anihamde: note that this was bumped while i was testing `sampo`. you may want to adjust this to a more appropriate version

PR #381, opened 2025-12-04
feat: add price endpoint back

anihamde:


packages/sessions-sdk-react/src/hooks/use-price.ts
```diff
@@ -9,18 +9,19 @@ export const usePrice = (mint: string) => {
...
+
+  const response = await fetch(priceUrl);
+  if (!response.ok) {
+    throw new Error(
+      `Failed to fetch price: ${response.status.toString()} ${response.statusText}`,
+    );
+  }
+
+  const price = priceSchema.parse(await response.json());
+  return price;
```
cprussin: Nit, but again no value in the assign then return, just return directly.

PR #395, opened 2025-12-05
chore: update changeset from changesets format to knope format

cprussin:


flake.nix
```diff
@@ -73,6 +73,19 @@
...
           doCheck = false;
         });
+
+        knope = final.rustPlatform.buildRustPackage (finalAttrs: {
+          pname = "knope";
+          version = "0.21.7";
+
+          src = final.fetchCrate {
+            inherit (finalAttrs) pname version;
+            hash = "sha256-ap4xip2QhsX1pqb2WZc60NfuSGMJxggD1XjMfC4C5ng=";
```
anihamde: i tried verifying these via running nix but kept running into some errors. i trust these are right

PR #438, opened 2025-12-18
feat: biome, baby!

benduran:
# What is this?

This PR swaps out both `eslint` and `prettier` for [Biome](https://biomejs.dev/).

**But why, I hear you ask?**

ESLint has a whole slew of issues and JavaScript tools that have been written with lower-level languages dramatically outperform ESLint, while offering a better developer experience.

## Changes in this PR
- Introduction of Biome and `biome.json` for configuring it
- Swap out eslint and prettier for the various `*:format` and `*:lint` commands to use biome
- Ran a `fix` pass over this repo to get the formatting and linting changes and suggestions
- Explicitly added `typescript` as a dependency to each package, as some subpackages were erroneously using a super old version of TypeScript that wasn't catching extremely valid build errors
- Speaking of build errors: once the line item, above, was addressed, I needed to update `sessions-provider.tsx`, `withdraw-page.tsx` and `send-token-page.tsx` to fix build and / or typing issues
- removed any `eslint-disable-*` comments in source code, as they're no longer needed
- updated any additional references to `eslint` in `package.json` or `turbo.json` files


packages/sessions-sdk-react/src/components/withdraw-page.tsx
```diff
@@ -260,11 +259,7 @@ const WithdrawFormImpl = (
     } catch {
       return;
     }
-  }, [
-    props.isLoading,
-    props.isLoading ? undefined : props.amount,
-    USDC.decimals,
-  ]);
```
alexcambose: just wanted to surface this in case this was intentional

packages/component-library/turbo.json
```diff
@@ -36,10 +36,10 @@
       "outputs": ["dist/theme.scss"]
     },
     "fix:lint": {
-      "dependsOn": ["fix:lint:eslint", "fix:lint:stylelint"],
+      "dependsOn": ["fix:lint:biome", "fix:lint:stylelint"],
```
cprussin: I think we can get rid of the `fix:lint:biome` task, the only reason why `fix:lint` got split into two tasks was because we previously had stylelint & eslint as two separate tools.  That distinction is now gone, so we can just use `fix:lint` and remove a lot of this pipeline from the turbo configs.

packages/sessions-sdk-react/src/components/session-provider.tsx
```diff
@@ -719,6 +712,14 @@ const useSessionState = ({
     ],
   );
 
+  /** refs */
+  const connectWalletRef = useRef(connectWallet);
+
+  /** effects */
+  useEffect(() => {
+    connectWalletRef.current = connectWallet;
+  });
```
cprussin: whoa what is going on here?  This seems like a hack around another issue and not a proper fix

packages/sessions-sdk-react/src/components/withdraw-page.tsx
cprussin: We should definitely not have `props` as a dependency of a memoization since props is unstable and is a new reference with each function call, this is effectively the same as not memoizing at all.

cprussin (changes requested):
Overall looking pretty good!

There are some issues with some of the changes to `send-token-page.tsx` and `session-provider.tsx`; I think these changes are actually incorrect and we need to figure out why they happened.

Additionally there's a few fairly useful rules I notice we're missing out on that I think we should ideally find a way to restore:

- `no-null` -- I don't care if we ban null or undefined, but we should ban one and stick to the other.  People are generally really bad at using both properly.  I've mostly stuck with `undefined` because it operates the way people think `null` should work, but I'd also be fine with switching to null, as long as we are banning one or the other.
- `no-process-env` -- We should only ever touch `process.env` in very specific places, and doing so elsewhere will cause bugs.  This rule helps ensure that doesn't happen accidentally.
- `no-console` -- I don't feel as strongly here but it's nice to help ensure we're intentional about what gets logged out where
- Some of the typescript rules seem to have gone away, e.g. ` @typescript-eslint/no-unsafe-argument` and `@typescript-eslint/no-explicit-any` in the change in https://github.com/fogo-foundation/fogo-sessions/pull/438/changes#diff-276fa2e9eca0f89f4df12e42cb3f3c6f3eb1af73cbb24651679fc44b85ea19f4L136. `@typescript-eslint/no-unused-vars`, `@typescript-eslint/no-unsafe-argument`, `@typescript-eslint/no-unsafe-member-access`, etc.  Given that `any` causes very subtle issues where code looks like it typechecks when in fact the typechecker is just turned off, I really think it's important to preserve rules that help us prevent `any` getting into the codebase unintentionally.
- `import/consistent-type-specifier-style` -- this rule was useful because it helps make sure the code is more tree-shakeable -- by forcing types to be imported separately, it means typescript can always strip all type imports out, see in particular the `import { type T } from 'T';` case in https://typescript-eslint.io/rules/consistent-type-imports/#comparison-with-importsnotusedasvalues--verbatimmodulesyntax

packages/sessions-sdk-react/src/components/session-provider.tsx
benduran: Yes, this is correct

there were a number of issues here, with the biggest offender being that the connect wallet callback was being used in another callback, but wasn’t set as a dep in that callback’s dep array, which could result in runtime bugs where the function is old and points to incorrect variables.

simply adding it as a dep array entry blew up the build, because it meant that connect wallet was being used before it was defined.

I made this latest Ref change to work around this, while ensuring connect wallet is always updated

cprussin: Yeah I really don't like this fix, I think we should either leave it as it is and add lint ignores or fix it properly.  This fix doesn't actually resolve the underlying issue and makes it even more hacky.

packages/sessions-sdk-react/src/components/send-token-page.tsx
```diff
@@ -314,18 +313,21 @@ const SendTokenPageImpl = ({
         maxSendAmount: bigint;
       }
   )) => {
-  const scannerShowing = !props.isLoading && props.scanner !== undefined;
+  const { isLoading } = props;
+  const amount = "amount" in props ? props.amount : "";
```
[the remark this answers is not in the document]
benduran: see previous comment about resetting all source files back to `origin/main`. let's discuss the biome settings, first, before I run another pass of `biome check --write` across the repo

services/paymaster/.sqlx/query-8adbc3bfb60a7802966a144aa94271828586b1f4a9f393e034993cb1405904e4.json
```diff
@@ -30,21 +30,13 @@
           "Custom": {
             "name": "network_environment",
             "kind": {
-              "Enum": [
-                "mainnet",
-                "testnet"
-              ]
+              "Enum": ["mainnet", "testnet"]
```
cprussin: We should probably ignore this entire directory from biome given this is managed by rust tooling & these files in particular are generated by sqlx

biome.json
```diff
@@ -0,0 +1,404 @@
...
+          "correctness": {
+            "noConstAssign": "off",
+            "noGlobalObjectCalls": "off",
+            "noInvalidBuiltinInstantiation": "off",
+            "noInvalidConstructorSuper": "off",
+            "noSetterReturn": "off",
+            "noUndeclaredVariables": "off",
+            "noUnreachable": "off",
+            "noUnreachableSuper": "off"
+          },
```
cprussin: These seem like really good rules, why turn them off?

cprussin (approved):
Nice, I'm very excited to ship this.  I have some minor nits / questions but as far as I can tell this is good to go.  I'm super pumped to see how this works out.

biome.json
benduran: for context, these were automatically set by the Biome ESLint migration tool. I agree, though. I've deleted all these so they fall back to their defaults

PR #450, opened 2025-12-23
feat: support fees in the typescript sdks

guibescos:


packages/sessions-sdk-ts/src/connection.ts
```diff
@@ -275,6 +311,53 @@ const buildTransaction = async (
   );
 };
 
+export const getDomainTollRecipientAddress = (domain: string) => {
```
cprussin: Any reason this is exported?

packages/sessions-sdk-ts/src/connection.ts
```diff
@@ -275,6 +311,53 @@ const buildTransaction = async (
...
+}: {
+  sessionKeyAddress: Address | undefined;
+  walletPublicKey: PublicKey;
+  domain: string;
+  feeMint: PublicKey;
+  feeAmount: BN;
+}) => {
+  if (feeAmount.eq(new BN(0)) || sessionKeyAddress === undefined) {
+    return [];
+  }
```
cprussin: Personal preference that I've noted before but I really dislike early returns and have tried to avoid them in the codebase here.  They are harder to trace and they mask the actual control flow of the code.  I prefer the control flow being explicit

packages/sessions-sdk-ts/src/connection.ts
```diff
@@ -275,6 +311,53 @@ const buildTransaction = async (
...
+  ).methods
+    .payToll(feeAmount)
+    .accountsPartial({
+      session: new PublicKey(sessionKeyAddress),
+      source: userTokenAccount,
+      destination: getAssociatedTokenAddressSync(feeMint, recipient, true),
+      mint: feeMint,
+    })
+    .instruction();
+  return [fromLegacyTransactionInstruction(instruction)];
```
cprussin: IMO this function should return an `Instruction | undefined`, not an `[Instruction]` with one or no items.

packages/sessions-sdk-ts/src/connection.ts
```diff
@@ -357,6 +440,34 @@ const getSponsor = async (
...
+    Parameters<typeof createSessionConnection>[0],
+    "paymaster" | "network"
+  >,
+  domain: string,
+  variation: string | undefined,
+  mint: PublicKey,
+) => {
+  if (variation === undefined) {
+    return new BN(0);
+  }
```
cprussin: Thoughts on just making variation mandatory in this release?

packages/sessions-sdk-ts/src/connection.ts
```diff
@@ -357,6 +440,34 @@ const getSponsor = async (
...
+    "/api/fee",
+    options.paymaster ?? DEFAULT_PAYMASTER[options.network],
+  );
+  url.searchParams.set("domain", domain);
+  url.searchParams.set("variation", variation);
+  url.searchParams.set("mint", mint.toBase58());
+  const response = await fetch(url);
+
+  if (response.status === 200) {
+    return new BN(z.string().parse(await response.text()));
```
cprussin: I don't think the zod parsing is necessary here, maybe I'm wrong but I think `response.text` returns `Promise<string>`, not `Promise<any>` right?

packages/sessions-sdk-ts/src/index.ts
```diff
@@ -738,7 +745,7 @@ export type Session = {
   sessionInfo: NonNullable<z.infer<typeof sessionInfoSchema>>;
 };
 
-const USDC_MINT = {
+export const USDC_MINT = {
```
cprussin: I don't think we should be exporting this and then importing it in `connection.ts` -- that introduces a circular dependency and some bundling tools have issues with circular dependencies.  It also adds `USDC_MINT` to the public exports of the package which I don't love since it doesn't seem like this package is the appropriate place to be importing token mints.

Rather than this, I think it would be better to just move the usdc mint addresses to a separate ts file and import from there in both places.

services/paymaster/src/api.rs
```diff
@@ -555,6 +555,62 @@ async fn sponsor_pubkey_handler(
     Ok(sponsors[sponsor_index].pubkey().to_string())
 }
 
+#[serde_as]
+#[derive(serde::Deserialize, utoipa::IntoParams)]
+#[serde(deny_unknown_fields)]
+#[into_params(parameter_in = Query)]
+struct FeeQuery {
+    #[serde(default)]
```
cprussin: Why are we enabling a default here?

cprussin (approved):

services/paymaster/src/api.rs
guibescos: We do the same in other places, if this is not provided it will try to get the domain from the origin header. I'd rather keeping it consistent for now.

packages/sessions-sdk-ts/src/connection.ts
```diff
@@ -357,6 +440,34 @@ const getSponsor = async (
...
+  variation: string | undefined,
+  mint: PublicKey,
+) => {
+  if (variation === undefined) {
+    return new BN(0);
+  }
+  const url = new URL(
+    "/api/fee",
+    options.paymaster ?? DEFAULT_PAYMASTER[options.network],
+  );
```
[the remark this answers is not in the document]
guibescos: let's tackle this later if someone requests it

PR #451, opened 2025-12-23
fix: tiltfile and service name of spans

guibescos:


services/paymaster/src/main.rs
```diff
@@ -27,7 +27,7 @@ async fn run_server(opts: cli::RunOptions) -> anyhow::Result<()> {
     let resource = opentelemetry_sdk::Resource::builder()
         .with_attributes(vec![opentelemetry::KeyValue::new(
             "service.name",
-            "paymaster-service",
+            format!("paymaster-service-{}", opts.network_environment),
```
guibescos: this is so we can distinguish spans from testnet and mainnet

PR #453, opened 2026-01-05
feat(paymaster): Auth and Apps UI

alexcambose:
Implementing more of the designs for the Auth and Apps screens.
Minor updates on data fetching query to also include the network_environment.

Still tbd with Greg on some of the details but will address those later.

https://github.com/user-attachments/assets/88d5ae28-3743-4bbe-aaef-b7e82d9c0135

### DIFF
--- a/apps/paymaster-admin/src/components/App/user-apps.tsx
+++ b/apps/paymaster-admin/src/components/App/user-apps.tsx
@@ -16,11 +16,10 @@ type AppCardProps =
       isLoading: true;
     };
 
-const AppCard = (props: AppCardProps) => {
-  if (props.isLoading) {
-    return <Skeleton className={styles.appCard} />;
-  }
-  return (
+const AppCard = (props: AppCardProps) =>
+  props.isLoading ? (
+    <Skeleton className={styles.appCard} />
+  ) : (
     <Card className={styles.appCard}>
       <div className={styles.appCardContent}>
         <h3 className={styles.appCardTitle}>{props.app.name}</h3>
@@ -35,7 +34,7 @@ const AppCard = (props: AppCardProps) => {
       </div>
     </Card>
   );
-};
+
 
 type UserAppsTitleProps =
   | {
--- a/apps/paymaster-admin/src/components/Auth/index.module.scss
+++ b/apps/paymaster-admin/src/components/Auth/index.module.scss
@@ -69,7 +69,7 @@
       background: linear-gradient(
         90deg,
         #{theme.color('border')} 0%,
-        #f6401b 47.46%,
+        #{theme.color('brand')} 47.46%,
         #{theme.color('border')} 98.72%
       );
       filter: blur(1px);
--- a/apps/paymaster-admin/src/components/App/index.tsx
+++ b/apps/paymaster-admin/src/components/App/index.tsx
@@ -62,14 +62,14 @@ const AppData = ({
     case StateType.NotLoaded: {
       return;
     }
+    case StateType.Loading: {
+      return <UserApps isLoading />;
+    }
     default: {
-      if (
-        (userData.type === StateType.Loaded && userData.data.type) ===
-        FetchUserDataStateType.NotFound
-      ) {
+      if (userData.data.type === FetchUserDataStateType.NotFound) {
         return <UserNotFound />;
       }
-      return <UserApps {...userData} />;
+      return <UserApps user={userData.data.user} />;
     }
   }
 };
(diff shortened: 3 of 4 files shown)


apps/paymaster-admin/src/app/api/auth/user-data/route.ts
```diff
@@ -1,9 +1,9 @@
-import { cacheTag, cacheLife } from "next/cache";
+import { cacheLife, cacheTag } from "next/cache";
 import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
-
 import { fetchUserPaymasterData } from "../../../../server/paymaster";
 
+// biome-ignore lint/suspicious/useAwait: "use cache" functions need to be async
```
cprussin: Easy way to resolve this without the lint ignore is `return await fetchUserPaymasterData` -- it used to be the case that `return await` was bad practice because it spawned a new microtask but that has long since no longer been the case and nowadays `return await` is preferred, see:

- https://eslint.org/docs/latest/rules/no-return-await#options
- https://typescript-eslint.io/rules/return-await/

apps/paymaster-admin/src/components/App/user-apps.tsx
```diff
@@ -1,18 +1,100 @@
...
+      app: App;
+      isLoading?: false;
+    }
+  | {
+      isLoading: true;
+    };
 
-export const UserApps = ({ user }: { user: User }) => {
+const AppCard = (props: AppCardProps) => {
+  if (props.isLoading) {
```
cprussin: nit: no reason to use early return

apps/paymaster-admin/src/components/Auth/index.module.scss
```diff
@@ -0,0 +1,91 @@
...
+    .connectWalletCardHighlightLine {
+      width: 204px;
+      height: 1px;
+      position: absolute;
+      top: 0;
+      left: theme.spacing(6);
+      background: linear-gradient(
+        90deg,
+        #{theme.color('border')} 0%,
+        #f6401b 47.46%,
```
cprussin: Should this intermediate color go in the themes?

apps/paymaster-admin/src/components/Footer/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+  <footer className={styles.footer}>
+    <div className={styles.footerContent}>
+      <div>
+        <Image src={fogoFooterImageSvg} alt="Fogo Paymaster" />
+        <p className={styles.footerText}>
+          Software is maintained by Dourolabs.xyz
+        </p>
+      </div>
+      <div>
+        <Link href="https://dourolabs.xyz" target="_blank">
```
cprussin: This should link to https://www.dourolabs.xyz/Paymaster-Terms-Of-Service.pdf

apps/paymaster-admin/src/components/Root/root.scss
```diff
@@ -0,0 +1,11 @@
...
+@use "@fogo/component-library/theme";
+
+html, body {
+  width: 100%;
+  height: 100%;
+  background-color: theme.color("background");
+  color: theme.color("foreground");
+  display: flex;
+  flex-direction: column;
+}
```
cprussin: Minor thing but all of this can be set in a css module in `Root` and then applied as classes to `<html>` and `<body>` in `Root/index.tsx` and I generally prefer that because it follows the conventions everywhere else in the app, so you don't have this weird thing where some styles are applied outside of css modules.

The only thing you have to put outside of a css module is the import to use `modern-normalize`.

apps/paymaster-admin/src/components/loading.tsx
```diff
@@ -1,7 +1,5 @@
+import { Skeleton } from "@fogo/component-library/Skeleton";
+
 export const PaymasterLoading = () => {
-  return (
-    <>
-      <div>Loading...</div>
-    </>
-  );
+  return <Skeleton />;
 };
```
cprussin: This could be `export { Skeleton as PaymasterLoading } from "@fogo/component-library/Skeleton"` (although I question why this module needs to exist tbh, could just import it like that in the route)

cprussin (approved):

apps/paymaster-admin/src/components/App/user-apps.tsx
```diff
@@ -1,18 +1,100 @@
...
+        </div>
+        <div className={styles.userApps}>
+          <AppCard isLoading />
+          <AppCard isLoading />
+          <AppCard isLoading />
+          <AppCard isLoading />
+        </div>
+      </>
+    );
+  }
```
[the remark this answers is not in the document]
alexcambose: nicer

PR #455, opened 2026-01-05
feat: session wrapping support

anihamde:
This PR adds support for native token wrapping into the paymaster. It does so by creating a new flag `requires_wrapped_native_tokens` which allows apps to denote a particular instruction in their constraints as needing to support the user wrapping native tokens. During the parsing of the config, the paymaster adds a 3-optional-instruction prefix consisting of:

1. system_program::session_wrap
2. associated_token_account::create_idempotent
3. token_program::sync_native

It also adds a single optional suffix token_program::close_account constraint to any instruction with `requires_wrapped_native_tokens` set to true.

This will enable use of WFOGO and other SPL tokens in the relevant instruction.

### DIFF
--- a/services/paymaster/src/constraint/config.rs
+++ b/services/paymaster/src/constraint/config.rs
@@ -24,15 +24,6 @@ impl From<TransactionVariation> for constraint::TransactionVariation {
     }
 }
 
-#[derive(Deserialize)]
-pub struct VariationOrderedInstructionConstraints {
-    pub name: String,
-    #[serde(default)]
-    pub instructions: Vec<InstructionConstraint>,
-    pub max_gas_spend: u64,
-    pub paymaster_fee_lamports: Option<u64>,
-}
-
 #[serde_as]
 #[derive(Deserialize)]
 pub struct InstructionConstraint {
@@ -64,6 +55,15 @@ impl From<InstructionConstraint> for constraint::InstructionConstraint {
             required,
         }
     }
+}
+
+#[derive(Deserialize)]
+pub struct VariationOrderedInstructionConstraints {
+    pub name: String,
+    #[serde(default)]
+    pub instructions: Vec<InstructionConstraint>,
+    pub max_gas_spend: u64,
+    pub paymaster_fee_lamports: Option<u64>,
 }
 
 impl From<VariationOrderedInstructionConstraints>


services/paymaster/src/constraint/mod.rs
```diff
@@ -258,6 +261,38 @@ pub struct InstructionConstraint {
...
+) -> Result<Vec<InstructionConstraint>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let constraints = Vec::<InstructionConstraint>::deserialize(deserializer)?;
+    let has_wrap_native = constraints.iter().any(|c| c.require_wrap_native);
+    let constraints_with_prefixes: Vec<InstructionConstraint> = constraints
+        .into_iter()
+        .flat_map(|base| {
+            if base.require_wrap_native {
```
cprussin: Should this actually expand into a pair of templates, one for the wrap variant and one for the unwrapped?  By not doing that we're requiring all configs that could interact with fogo or other spl tokens to have to be duplicated in all configs right?

How about something like having two separate options `require_wrap_native` OR `optionally_wrap_native`, and then the team can just enable `require_wrap_native` if the tx should only be sponsored if wrapping, and `optionally_wrap_native` if wrapping is supported but not required?

anihamde: `require_wrap_native` by default makes all the prefix and suffix instructions optional. the require is just meant to indicate that the native tokens need to be in wrapped form for the instruction to work. can fix the naming if you think this is confusing

cprussin: ah that makes sense, I do think the naming is a bit confusing but also I could imagine a use case where an app _only_ supports native tokens (e.g. liquid staking) and thus wants to make their paymaster reject any attempt to sponsor txns that don't have the wrap

anihamde: maybe, but for now i think it's fine to keep as is for simplicity? anyways, any txs that ppl submit without the wrapping should fail if they don't end up with WFOGO. and there could be cases where a user already has WFOGO and doesn't need to wrap them. after all, the frontend ultimately gets to choose whether to include the wrapping/unwrapping ixs

services/paymaster/src/constraint/config.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub accounts: Vec<AccountConstraint>,
+    #[serde(default)]
+    pub data: Vec<DataConstraint>,
+    pub required: bool,
+    #[serde(default)]
+    pub requires_wrapped_native_tokens: bool,
+}
+
+impl From<InstructionConstraint> for constraint::InstructionConstraint {
+    fn from(config: InstructionConstraint) -> Self {
```
guibescos: suggestion: I like this syntax lately that @cprussin uses a lot 
```rust
impl From<InstructionConstraint> for constraint::InstructionConstraint {
    fn from(InstructionConstraint {
        program,
        accounts,
        data,
        required,
        requires_wrapped_native_tokens: _,
    }: InstructionConstraint) -> Self {
        constraint::InstructionConstraint {
            program,
            accounts,
            data,
            required,
        }
    }
}
```

PR #457, opened 2026-01-06
Moving reset back to the component library

alexcambose:
Moving back the reset to the component library instead of passing a reset class from the react-sdk, as it's very difficult to use reset from an external class.

Also this PR got way bigger since I had to resolve some linting issues from biome.

for reference this is the initial PR that moved reset from component library to react-sdk (and didn't work) https://github.com/fogo-foundation/fogo-sessions/pull/440

### DIFF
--- a/packages/sessions-sdk-react/integration-tests/index.test.ts
+++ b/packages/sessions-sdk-react/integration-tests/index.test.ts
@@ -49,7 +49,7 @@ const runTestApp = async (port: number) => {
 
   return {
     page,
-    closeApp: async () => {
+    closeApp: () => {
       next.kill();
       return Promise.all([
         browser.close(),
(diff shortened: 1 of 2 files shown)


biome.json
```diff
@@ -338,31 +326,33 @@
...
+        "globals": [
+          "describe",
+          "it",
+          "test",
+          "expect",
+          "beforeEach",
+          "afterEach",
+          "beforeAll",
+          "afterAll"
+        ]
```
alexcambose: @benduran added this since biome was throwing errors like `The expect variable is undeclared.`

benduran: appreciate that. this will help others that write tests using jest

benduran (approved):
LGTM

cprussin (approved):
Gave it a quick scroll through, it all makes sense to me.  Didn't look too in depth because the change is mostly procedural but lmk if there's anything you feel like you need a closer eye on.

PR #463, opened 2026-01-06
feat: support native token intent transfers

cprussin:


programs/intent-transfer/src/intrachain/processor/send_native.rs
```diff
@@ -0,0 +1,178 @@
...
+use anchor_spl::{
+    associated_token::AssociatedToken,
+    token::{spl_token::try_ui_amount_into_amount, Mint, Token, TokenAccount},
+};
+use chain_id::ChainId;
+use solana_intents::{Intent, SymbolOrMint};
+
+const NONCE_SEED: &[u8] = b"nonce";
+const FOGO_DECIMALS: u8 = 9;
+const INTENT_TRANSFER_DISCRIMINATOR: u32 = 4_000_001;
```
anihamde: can we name this like `SYSTEM_PROGRAM_INTENT_TRANSFER_DISCRIMINATOR` to be clear?

programs/intent-transfer/src/intrachain/processor/send_native.rs
```diff
@@ -0,0 +1,178 @@
...
+        require_keys_eq!(
+            recipient,
+            destination.key(),
+            IntentTransferError::RecipientMismatch
+        );
+
+        verify_and_update_nonce(nonce, new_nonce)?;
+
+        let amount = try_ui_amount_into_amount(amount, FOGO_DECIMALS)?;
+        let mut data = Vec::with_capacity(16);
```
anihamde: you can construct this without making `data` `mut`

programs/intent-transfer/src/intrachain/processor/send_native.rs
```diff
@@ -0,0 +1,178 @@
...
+                },
+            signer,
+        } = Intent::load(sysvar_instructions.as_ref())
+            .map_err(Into::<IntentTransferError>::into)?;
+
+        if chain_id.chain_id != expected_chain_id {
+            return err!(IntentTransferError::ChainIdMismatch);
+        }
+
+        if symbol_or_mint != SymbolOrMint::Symbol(String::from("FOGO")) {
```
anihamde: what about if `== SymbolOrMint::Mint("So111....112")`?

packages/sessions-sdk-react/src/components/session-limits.tsx
```diff
@@ -254,7 +254,9 @@ const LoadedTokenLimits = ({
       tokenAccountData === undefined
         ? []
         : [
-            ...tokenAccountData.tokensInWallet,
+            ...tokenAccountData.tokensInWallet.filter(
+              (token) => !token.isNative,
```
anihamde: maybe a comment explaining why we exclude the wrapped native mint?

packages/sessions-sdk-react/src/hooks/use-token-account-data.ts
```diff
@@ -61,15 +63,40 @@ const getTokenAccounts = async (
...
+      ...splAccounts
+        .filter(({ amountInWallet }) => amountInWallet !== 0n)
+        .map(({ mint, amountInWallet, decimals }) => ({
+          isNative: false as const,
+          mint: new PublicKey(mint),
+          amountInWallet,
+          decimals,
+          ...metadata[mint],
+        }))
+        .toSorted((a, b) => {
```
anihamde: i don't recall if we directly show the sorted `tokenList` in `token-list.tsx`, but if so, maybe we should sort the native token balance first?

packages/sessions-sdk-react/src/components/session-limits.tsx
cprussin: We do not exclude the wrapped native mint from the session limits, we exclude the actual native token itself from session limits (which makes sense given the only thing you could do with the native token is either transfer it or wrap it and then work with the wrapped token, and you can apply limits to the wrapped token).

I can add a comment explaining this.

programs/intent-transfer/src/intrachain/processor/send_native.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub nonce: Account<'info, Nonce>,
+
+    #[account(mut)]
+    pub sponsor: Signer<'info>,
+
+    #[account(mut, token::mint = fee_mint, token::authority = source)]
+    pub fee_source: Account<'info, TokenAccount>,
+
+    #[account(init_if_needed, payer = sponsor, associated_token::mint = fee_mint, associated_token::authority = sponsor)]
+    pub fee_destination: Account<'info, TokenAccount>,
```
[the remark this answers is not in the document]
cprussin: yes I agree we need to do this, although maybe we can make native transfers free since there's no rent vulnerability?

PR #465, opened 2026-01-07
feat(component-library): add skeleton component

alexcambose:
Screenshot 2026-01-07 at 13 52 05

Add a skeleton component - taken from pyth-crosschain mostly.
Also added a missing reset.

### DIFF
--- a/packages/component-library/src/Skeleton/index.tsx
+++ b/packages/component-library/src/Skeleton/index.tsx
@@ -17,7 +17,7 @@ export const Skeleton = ({
   ...props
 }: Props) => (
   <span
-    data-fill-width={width === undefined ? "" : undefined}
+    data-fill-width={width === undefined}
     {...(width &&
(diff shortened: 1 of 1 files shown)


packages/component-library/src/Skeleton/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const Skeleton = ({
+  className,
+  label,
+  width,
+  fill,
+  ...props
+}: Props) => (
+  <span
+    data-fill-width={width === undefined ? "" : undefined}
```
benduran: nit: if you have `true` as the value, the `data-fill-width` attribute will get written out as `data-fill-width="true"`. this means you can write this statement like this:
```
data-fill-width={width === undefined}
```

cleans things up a little 🙂

alexcambose: yep, ty!

PR #494, opened 2026-01-08
feat(paymaster-admin): add i18n-provider

alexcambose:


packages/component-library/src/I18nProvider/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+      isSSR
+        ? DEFAULT_LOCALE
+        : (globalThis.navigator.languages.find((locale) => {
+            const language = parse(locale).language;
+            return (
+              language !== undefined &&
+              language !== null &&
+              SUPPORTED_LANGUAGES.has(language)
+            );
+          }) ?? DEFAULT_LOCALE),
```
cprussin: I think we can actually just hardcode to `en-US` for now right?  We aren't going to localize anything else...

Originally when I had this in crosschain the idea was to localize things like number & date formats, but we found that to be very confusing to users since only parts of the UI were localized while the rest of the UI wasn't.  I nowadays think it's better to just stick to `en-US` until we're ready to actually localize fully.  What are your thoughts?

alexcambose: yeah I think that's a good point. We can think about this when we do think about localization

PR #515, opened 2026-01-12
feat(paymaster-admin): variations screen design

alexcambose:
Adding the variations screen. 
To reduce the scope here some of the changes are just the barebones structure of how it would look and the design, there's no validation or logic for interacting and changing the variation properties.
Also for the code editor I used https://github.com/securingsincity/react-ace which seems to work nicely.

https://github.com/user-attachments/assets/d3b37e18-0890-438c-9e72-008a38c41029

### DIFF
--- a/apps/paymaster-admin/package.json
+++ b/apps/paymaster-admin/package.json
@@ -30,7 +30,9 @@
     "react-aria": "catalog:",
     "react-aria-components": "catalog:",
     "react-dom": "catalog:",
+    "smol-toml": "catalog:",
     "swr": "catalog:",
+    "uuid": "catalog:",
     "zod": "catalog:"
   },
   "devDependencies": {
--- a/apps/paymaster-admin/src/components/ConfirmModal/index.module.scss
+++ b/apps/paymaster-admin/src/components/ConfirmModal/index.module.scss
@@ -1,0 +1,42 @@
+@use '@fogo/component-library/theme';
+
+.confirmModalHeader {
+  background-color: theme.color('background');
+  padding: theme.spacing(6);
+  border-bottom: 1px solid theme.color('border');
+  display: flex;
+  flex-direction: column;
+  gap: theme.spacing(4);
+
+  .confirmModalHeaderAltText {
+    @include theme.text('sm', 'normal');
+    color: theme.color('muted');
+  }
+
+  .confirmModalHeaderTitle {
+    @include theme.text('3xl', 'normal');
+    color: theme.color('foreground');
+  }
+
+  .confirmModalHeaderSubtitle {
+    @include theme.text('base', 'normal');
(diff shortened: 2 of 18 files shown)


apps/paymaster-admin/src/components/BreadcrumbNav/index.tsx
```diff
@@ -12,24 +12,84 @@ type Item =
   | { label: string; href?: string; isLoading?: false }
   | { isLoading: true };
 
+type TitleProps = {
+  title?: string | undefined;
+  titleLoading?: boolean | undefined;
+};
+
+type BreadcrumbNav = TitleProps & {
```
anihamde: rename the type to distinguish it from the component

apps/paymaster-admin/src/components/Variation/app-variations.tsx
```diff
@@ -0,0 +1,56 @@
...
+    };
+
+export const AppVariation = (props: AppVariationProps) => {
+  return (
+    <>
+      <BreadcrumbNav
+        items={[
+          props.isLoading
+            ? { isLoading: true }
+            : { label: props.app.name, href: `/apps/${props.app.id}` },
```
anihamde: this can also include the first Apps page as an href

apps/paymaster-admin/src/components/BreadcrumbNav/index.tsx
```diff
@@ -12,24 +12,84 @@ type Item =
...
   const router = useRouter();
   const handleBackClick = useCallback(() => {
-    const lastItem = items.at(-1);
-    if (!lastItem?.isLoading) {
-      const lastLink = lastItem?.href;
-      if (lastLink) {
-        router.push(lastLink);
-      }
+    const lastItemWithHref = items.findLast(
+      (item) => !item.isLoading && item.href,
```
anihamde: should loading status affect the item that back routes to? i'd prefer back arrow to just not be available/functional until all the items (or at least the last one with href) have loaded fully

apps/paymaster-admin/src/components/Variation/add-variation-button.tsx
```diff
@@ -0,0 +1,39 @@
+import { Button } from "@fogo/component-library/Button";
+import { ModalDialog } from "@fogo/component-library/ModalDialog";
+import { StackPlusIcon } from "@phosphor-icons/react/dist/ssr";
+import { useCallback, useState } from "react";
+
+const AddVariationModal = ({
```
anihamde: perhaps instead of a modal this should just create a new empty row in the variations that needs to be populated?

alexcambose: actually yeah let's remove this for now and just use the inline editing

PR #519, opened 2026-01-13
chore: clean parsing

guibescos:
I'd like to further simplify parsing by enforcing that for a multiline value in the intent message, it may only be a list with each item starting by `-`.

This solves some ambiguous cases there are now like: `foo: \nbar` is interpreted as `{key: "foo", value: ""}` but `foo:\nbar` is interpreted as `{key: "foo", value: "bar"}`

### DIFF
--- a/packages/solana-intents/src/key_value.rs
+++ b/packages/solana-intents/src/key_value.rs
@@ -2,7 +2,7 @@ use nom::{
     branch::alt,
     bytes::complete::{tag, take_while1},
     character::complete::{char, line_ending, not_line_ending},
-    combinator::{eof, map, map_opt, opt, recognize},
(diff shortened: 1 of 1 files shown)


packages/solana-intents/src/key_value.rs
```diff
@@ -58,16 +58,11 @@ where
...
-                        peek(alt((
-                            preceded(
-                                line_ending,
-                                take_while1(|c: <I as Input>::Item| c.as_char() != '-'),
-                            ),
-                            eof,
-                        ))),
-                    )),
+                    recognize(many0(delimited(
+                        opt(line_ending),
```
cprussin: should this be `opt(alt([line_ending, eof]))` like we have on line 58?

PR #525, opened 2026-01-15
fix: enforce type imports / exports being separate from code imports / exports. also fixed one occurrence of this :)

benduran:
## Screenshot of error

image


cprussin (approved):
Nice, thank you.  Can you also add a changeset with `knope document-change` please?

PR #535, opened 2026-01-19
chore: unwrap lint

guibescos:


.github/workflows/cargo-clippy.yaml
```diff
@@ -19,3 +19,7 @@ jobs:
         env:
           SQLX_OFFLINE: true
         run: cargo clippy --tests -- -D warnings
+      - name: Clippy tests
+        env:
+          SQLX_OFFLINE: true
+        run: cargo clippy --tests -- -D warnings -A clippy::unwrap_used
```
guibescos: I think we should allow unwrap in tests, and this is the way that I found to do it.
We can also add other lints here that we think are too strict for tests.

PR #536, opened 2026-01-20
chore: improve data constraints

guibescos:
This PR improves data constraints:
- removes the `data_type` key since it can be inferred from `constraint`
- moves a lot of runtime checks into the config loading flow. This means errors in the constraints get caught earlier instead of leading to obscure runtime errors.

It completely compatible with the current config format and just improves the internal types.

### DIFF
--- a/.changeset/make_the_data_type_field_in_data_constraints_optional.md
+++ b/.changeset/make_the_data_type_field_in_data_constraints_optional.md
@@ -1,0 +1,5 @@
+---
+fogo-paymaster: patch
+---
+
+# Make the data_type field in data constraints optional


services/paymaster/src/constraint/config.rs
```diff
@@ -34,6 +39,235 @@ impl TransactionVariation {
...
+            DataConstraintSpecification::EqualTo(values) => parse_equal_values(values, true),
+            DataConstraintSpecification::Neq(values) => parse_equal_values(values, false),
+        }
+    }
+}
+
+fn parse_equal_values(
+    values: Vec<DataValue>,
+    is_equal: bool,
+) -> Result<ParsedDataConstraintSpecification, anyhow::Error> {
```
anihamde: `anyhow::Result`

services/paymaster/src/constraint/mod.rs
```diff
@@ -470,80 +463,123 @@ impl DataConstraint {
...
+}
+
+pub enum IntegerConstraint<T> {
+    LessThan(T),
+    GreaterThan(T),
+    EqualTo(Vec<T>),
+    Neq(Vec<T>),
+}
+
+pub enum ScalarConstraint<T> {
```
anihamde: maybe `NonNumericalScalarConstraint`? maybe too verbose

PR #542, opened 2026-01-21
feat: paymaster fee swap

anihamde:
This PR adds functionality to the paymaster to enable swapping out of collected fee tokens into FOGO to self-fund operations. The swaps take place on Valiant and are performed randomly to mitigate predictability of market impact.

Metrics relating to the swaps will be added in a follow-up PR.

### DIFF
--- a/services/paymaster/src/api.rs
+++ b/services/paymaster/src/api.rs
@@ -462,8 +462,8 @@ async fn sponsor_and_send_handler(
         // This ensures that any logs/traces from the spawned task are associated with the original request.
         let span = tracing::Span::current();
 
-        let state_clone = Arc::clone(&state);
-        tokio::spawn(
+        tokio::spawn({
+            let state = Arc::clone(&state);
             async move {
                 match fetch_transaction_cost_details(
                     &state.chain_index.rpc,
@@ -491,22 +491,25 @@ async fn sponsor_and_send_handler(
                     }
                 }
             }
-            .instrument(span),
-        );
+            .instrument(span)
+        });
 
         let transaction_sponsor_clone = transaction_sponsor.clone();
 
-        tokio::spawn(async move {
-            if let Some(valiant_client) = &state_clone.valiant_client {
-                valiant_client
-                    .swap_tokens_probabilistic(
-                        &swap_into_fogo,
-                        &transaction_sponsor_clone,
-                        &state_clone.chain_index.rpc,
-                    )
-                    .await;
-            }
-        });
+        if !swap_into_fogo.is_empty() {
+            tokio::spawn(async move {
+                if let Some(valiant_client) = &state.valiant_client {
+                    valiant_client
+                        .swap_tokens_probabilistic(
+                            &swap_into_fogo,
+                            &transaction_sponsor_clone,
+                            &state.chain_index.rpc,
+                            &state.rpc_sub,
+                        )
+                        .await;
+                }
+            });
+        }
     }
 
     Ok(Json(confirmation_result.into()))
--- a/services/paymaster/src/cli.rs
+++ b/services/paymaster/src/cli.rs
@@ -90,7 +90,7 @@ pub struct RunOptions {
     #[arg(long, env = "VALIANT_API_KEY")]
     pub valiant_api_key: Option<String>,
 
-    #[arg(long)]
+    #[arg(long, env = "VALIANT_OVERRIDE_URL")]
     pub valiant_override_url: Option<String>,
 }
 
--- a/services/paymaster/src/swap.rs
+++ b/services/paymaster/src/swap.rs
@@ -14,12 +14,12 @@ use std::str::FromStr;
 use url::{form_urlencoded, Url};
 
 use crate::{
-    api::ValiantClient,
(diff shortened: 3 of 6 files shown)


services/paymaster/src/swap.rs
```diff
@@ -0,0 +1,289 @@
...
+    output_mint: Pubkey,
+    input_amount: u64,
+    is_exact_in: bool,
+    direct_only: bool,
+}
+
+#[serde_as]
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[allow(dead_code)]
```
anihamde: this is bc there are fields in the TwoHopQuoteResponse that we don't need to use. we could also get rid of the fields that we don't care about and serde will ignore them. i'm leaning toward that, but open to feedback

services/paymaster/src/api.rs
```diff
@@ -481,6 +490,21 @@ async fn sponsor_and_send_handler(
...
+
+        let transaction_sponsor_clone = transaction_sponsor.clone();
+
+        tokio::spawn(
+            async move {
+                if let Some(valiant_client) = &state_clone.valiant_client {
+                    valiant_client.swap_tokens_probabilistic(
+                        &swap_into_fogo,
+                        &transaction_sponsor_clone,
+                        &state_clone.chain_index.rpc,
```
anihamde: we can also put a reference to the rpc_client inside the valiant_client

services/paymaster/src/cli.rs
```diff
@@ -85,6 +86,12 @@ pub struct RunOptions {
 
     #[arg(long, env = "DB_REFRESH_INTERVAL_SECONDS", default_value = "10")]
     pub db_refresh_interval_seconds: u64,
+
+    #[arg(long, env = "VALIANT_API_KEY")]
+    pub valiant_api_key: Option<String>,
+
+    #[arg(long)]
+    pub valiant_override_url: Option<String>,
```
guibescos: maybe needs an env variable for consistency

services/paymaster/src/api.rs
```diff
@@ -481,6 +493,20 @@ async fn sponsor_and_send_handler(
             }
             .instrument(span),
         );
+
+        let transaction_sponsor_clone = transaction_sponsor.clone();
+
+        tokio::spawn(async move {
+            if let Some(valiant_client) = &state_clone.valiant_client {
```
guibescos: mmmmm maybe we shouldn't spawn the thread if:
- spaw_into_fogo is empty
- valiant_client is None

services/paymaster/src/api.rs
```diff
@@ -481,6 +493,20 @@ async fn sponsor_and_send_handler(
...
+                valiant_client
+                    .swap_tokens_probabilistic(
```
guibescos: handle the errors?

services/paymaster/src/cli.rs
```diff
@@ -85,6 +86,12 @@ pub struct RunOptions {
 
     #[arg(long, env = "DB_REFRESH_INTERVAL_SECONDS", default_value = "10")]
     pub db_refresh_interval_seconds: u64,
+
+    #[arg(long, env = "VALIANT_API_KEY")]
+    pub valiant_api_key: Option<String>,
```
guibescos: it was slightly unintuitive to me that there's no client if you don't pass an API key.
I would've though it this is None then we have a client but without API key.

services/paymaster/src/api.rs
```diff
@@ -100,12 +103,18 @@ impl PubsubClientWithReconnect {
     }
 }
 
+pub struct ValiantClient {
```
guibescos: does this belong in swap.rs.
I find it weird that the `impl ValiantClient` block is in swap.rs

services/paymaster/src/swap.rs
guibescos: Yeah I think both options make sense, having all fields make sense too as it serves as a kind of reference it we want to update the logic to use some other fields.

services/paymaster/src/api.rs
guibescos: This seems fine to me

guibescos (approved):

services/paymaster/src/api.rs
anihamde: if mint_swap_rates is empty sure, i'll add that. for the valiant_client being Some, there are some challenges around passing in the reference valiant_client into the async move block, so i decided to do it this way.

services/paymaster/src/swap.rs
```diff
@@ -0,0 +1,317 @@
...
+struct TwoHopQuoteResponse {
+    #[serde_as(as = "DisplayFromStr")]
```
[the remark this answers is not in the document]
anihamde: it's needed here bc the quote response has the output amount as a string (e.g. "100000")

services/paymaster/src/api.rs
anihamde: yes for sure. will do inside `swap_tokens_probabilistic`, but will do once i get to metrics

